PR #31, opened 2017-12-28
support python3 run_data.sh

rgtjf:
support python3 for data preprocess

### DIFF
--- a/data/WikiQA/filter_query.py
+++ b/data/WikiQA/filter_query.py
@@ -1,11 +1,16 @@
-#!/usr/bin/env python
 # coding: utf-8
+"""
+filter queries which have no right or wrong answers
+Given dev.txt, dev.ref, dev-filtered.ref
+Output dev-filtered.ref
+"""
+from __future__ import print_function
 
 import os
 import sys
 
-
 if __name__ == '__main__':
+
     basedir = './WikiQACorpus/'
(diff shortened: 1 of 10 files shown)


uduse:
just wondering, is there a reason of using `codecs.open` instead of `open`?

faneshion:
It's very nice for contributing your code @rgtjf.  I have the same question with uduse, why replace open with codecs.open?

rgtjf:
python2:
   when using `open`, it may happen `UnicodeDecodeError` . Therefore, we could use codecs to encode `str` into `utf8`

python3:
  it can use `open` to set the file encoding. Moreover, `codecs.open`  also works.

Thus for Compatibility, I use `codecs.open` instead of `open`.

ref:
1. https://zhidao.baidu.com/question/1823800427851275948.html
1. http://python.jobbole.com/81244/

rgtjf:
I will report the results soon

rgtjf:
I run ARC-I twice, the results are following:

| Models | NDCG@3   | NDCG@5   | MAP      |
|--------|----------|----------|----------|
| ARC-I  | 0.5680   | 0.6317   | 0.5870   |
| Run1   | 0.562959 | 0.626947 | 0.580135 |
| Run2   | 0.573168 | 0.637427 | 0.589967 |

Note:
1. I obtain `18677` tokens in file `word_dict.txt`, while the origin is `18670`. To run the code, I also modify the `vocab_size`  in the config file.

PR #55, opened 2018-02-20
Fix import errors in matchzoo/models/knrm.py

EikeDehling:
Thanks for making MatchZoo !

I was running one of the examples:

python matchzoo/main.py --phase train --model_file examples/toy_example/config/knrm_ranking.config

That example failed because of import errors in matchzoo/models/knrm.py

I fixed the import errors in this PR - then the example works properly


bwanglzu:
I'll close this PR at the moment, thanks!

PR #58, opened 2018-02-26
a bug in a function not been used

ChangQF:
It seems that this function is useless, but I want to fix it because my ocd. :)


changlinzhang:
@ChangQF 
Hi, I am a user of MatchZoo. Thanks for your work.
I wonder if this error will result in low performance results as #99 
I met the problem but dont know its reason and how to fix it.

bwanglzu:
Anyone could figure it out?

PR #81, opened 2018-04-17
check download file

rockyzhengwu:
1. I run the script `run_data.sh` many times , in order to avoid download every time .
2. if Downloading files failed at some step , the next time will not download all .


bwanglzu:
this is a good one actually

Issue #103, opened 2018-05-17
Published results not getting reproduced


aneesh-joshi:
Hi,
I am trying to reproduce the ndcg scores published on the readme
However, I am not getting similar numbers.

I run the `run_xxxx.sh` functions provided in `examples/wikiqa/`
On the prediction task, I get ndcg@3 values around 0.3 AS compared to the 0.5 to 0.6 range published.

Could I be doing something wrong?
Isn't running the network with `--predict` the correct way of getting there?

Example output:
`[05-16-2018 19:19:00]   [Predict] @ predict [Predict] results:  ndcg@3=0.379165 ndcg@5=0.451163 map=0.416128` for cdssm

and

`[05-17-2018 18:24:32]   [Predict] @ predict [Predict] results:  ndcg@3=0.341159 ndcg@5=0.415373 map=0.391363` for mvlstm

aneesh-joshi:
@faneshion

bwanglzu:
Same issue [here](https://github.com/faneshion/MatchZoo/issues/99)

yangliuy:
@aneesh-joshi @bwanglzu See my reply [here] (https://github.com/faneshion/MatchZoo/issues/99)

aneesh-joshi:
discussion going on in 
https://github.com/faneshion/MatchZoo/issues/99

So, closing

Issue #80, opened 2018-04-15
reproduction of the results


Chunlinx:
Hi,
When I run the code on MSRP dataset ,I only got the accuracy result 66% not the same as the result on the paper 76%, what the basic parameter setup in your experiments?

yangliuy:
For PI, we have incorporated results on QuoraQP (Quora Question Pairs) data insteads of MSRP in the readme file of MatchZoo. Which model did you refer to in your question ?

bwanglzu:
Please check #99 .

PR #122, opened 2018-05-25
wikiqa config lr

faneshion:
#80 #99 #103


bwanglzu:
I do not have input, let's wait for @yangliuy review.

Issue #69, opened 2018-04-03
error about matchpyramid on wikiqa


ckqsars:
I have succeeded train with model drmm. However when I want to use matchpyramid to train the model, it reply the error

/DynamicMaxPooling.py:37: RuntimeWarning: divide by zero encountered in divide
  stride2 = 1.0 * max_len2 / len2_one
py:36: RuntimeWarning: divide by zero encountered in divide
  stride1 = 1.0 * max_len1 / len1_one

I know the means of the error, but how should i do for it

Thank you

thiziri:
What's the configuration that you use in the .config file for MatchPyramid, check the parameters: text1_maxlen and text2_maxlen.

ckqsars:
The code I use is
 'python matchzoo/main.py --phase train --model_file examples/wikiqa/config/matchpyramid_wikiqa.config'  the configuration is your matchpyamid_wikiqa.config 

and I record the 'max_len1 len1_one max2_len len2_one' and I found this record
max_len1,len1_one,max2_len,len2_one,
10,6,40,0
the one of the text2 length is zeros. Is this data cause the error?

thiziri:
text1_maxlen and text2_maxlen should give the maximum length of text1 and text2 respectively, so they must be different from 0

bwanglzu:
Possibly a bug, need someone to look into this.

bwanglzu:
@ckqsars @thiziri Maybe it's  a good point to contribute some code to MatchZoo　:) to handle `ZeroDivisionError`.

aneesh-joshi:
Having the same problem

bwanglzu:
`dynamic_pooling_index` is defined as:

```python
@staticmethod
def dynamic_pooling_index(len1, len2, max_len1, max_len2):
    def dpool_index_(batch_idx, len1_one, len2_one, max_len1, max_len2):
        stride1 = 1.0 * max_len1 / len1_one
        stride2 = 1.0 * max_len2 / len2_one
        idx1_one = [int(i/stride1) for i in range(max_len1)]
        idx2_one = [int(i/stride2) for i in range(max_len2)]
        mesh1, mesh2 = np.meshgrid(idx1_one, idx2_one)
        index_one = np.transpose(np.stack([np.ones(mesh1.shape) * batch_idx, mesh1, mesh2]), (2,1,0))
        return index_one
    index = []
    for i in range(len(len1)):
[...]

PR #140, opened 2018-05-29
Feature/dynamic pool div 0

pl8787:
#69 

1. Solve `DivideBy0Error` in `DynamicMaxPooling`.
2. Add `.coverage` to git ignore.


bwanglzu:
thanks

matchzoo/layers/DynamicMaxPooling.py
```diff
@@ -31,16 +43,32 @@ def compute_output_shape(self, input_shape):
         return (None, self.psize1, self.psize2, input_shape_one[3])
 
     @staticmethod
-    def dynamic_pooling_index(len1, len2, max_len1, max_len2):
+    def dynamic_pooling_index(len1, len2, max_len1, max_len2, 
+                              compress_ratio1 = 1, compress_ratio2 = 1):
```
bwanglzu: This dosen't need to be changed, but for function arguments, we do not need a blank:

```python
def dynamic_pooling_index(len1,
len2,
...
compress_ratio1=1,
compress_ratio2=1
```

pl8787: OK

PR #145, opened 2018-05-29
Add model in config to prevent overwriting of predict.test.wikiqa.txt

aneesh-joshi:
Currently, every time `main.py` is run with `--phase predict`, the TREC format file is overwritten
So, if you predict with dssm and then cdssm, the cdssm prediction is overwritten.

To prevent this, I have added the names of the models.
We could go a step further and add time but I haven't done that.


aneesh-joshi:
On an unrelated note,
I am currently running benchmarks and wanted to understand the role of the `toy_example/config/` configs like `xxxx_classify.config` and `xxxx_ranking.config`

What are these and how are they different from the configs in `examples/wikiqa`?
Should I include them in my benchmark?

bwanglzu:
@aneesh-joshi Let's wait for the reviewers, I'm not familiar with this part as well..

aneesh-joshi:
@bwanglzu 
Do you think it would be a good idea to add an Issue for "Suggestions for 2.0" or "Wishlist/Todo for 2.0" which could be added to the Project tab?

bwanglzu:
@aneesh-joshi Yes, that's a good one, can you create them? Then I'll label them.

Issue #156, opened 2018-05-31
engine\base_model


bwanglzu:
Design `base_model.py` for MZ 2.0, serve as the abstract base class for all models.

PR #159, opened 2018-06-02
Feature/base model

uduse:
#156


matchzoo/models/dense_baseline_model.py
```diff
@@ -0,0 +1,34 @@
...
+        return params
+
+    def build(self):
+        x_in = [keras.layers.Input(shape)
+                for shape in self._params['input_shapes']]
+
+        x = keras.layers.concatenate(x_in)
+        x = keras.layers.Dense(
+                self._params.num_dense_units, activation='relu')(x)
+        x_out = keras.layers.Dense(1, activation='sigmoid')(x)
```
bwanglzu: activation sould in `params`

matchzoo/tasks/classification.py
```diff
@@ -0,0 +1,34 @@
...
+from matchzoo import engine
+
+
+class Classification(engine.BaseTask):
+    """Classification task."""
+
+    def __init__(self, num_classes: int = 2):
+        super().__init__()
+        if not isinstance(num_classes, int):
+            raise TypeError
```
bwanglzu:
```python
import this

Explicit is better than implicit.
```

matchzoo/tasks/classification.py
```diff
@@ -0,0 +1,34 @@
...
+        if num_classes < 2:
+            raise ValueError
```
bwanglzu: Explicit is better than implicit.

matchzoo/tasks/ranking.py
```diff
@@ -1,5 +1,7 @@
 """Ranking task."""
 
+import keras
```
bwanglzu: do you need to import entire library?

matchzoo/engine/base_model.py
```diff
@@ -1,207 +1,107 @@
...
-        self._num_hidden_layers = None
+import typing
+import pickle
+from pathlib import Path
 
-    @property
-    def name(self):
-        """Get model name."""
-        return self._name
+import keras
```
bwanglzu: dependency

matchzoo/engine/base_model.py
```diff
@@ -1,207 +1,107 @@
...
-        conf.update(self._fixed_hyper_parameters)
-        return conf
+    def backend(self) -> keras.models.Model:
+        return self._backend
 
     @abc.abstractmethod
-    def _build(self):
-        """Build model, each sub class need to impelemnt this method.
+    def build(self):
+        """"""
```
bwanglzu: so basically, you didn't write any docstring  and delete my docstring.

matchzoo/models/dense_baseline_model.py
uduse: Not in this case. This is the situation where the activation function is one of the fixed params, and got hard coded in to the build. I wanted to keep the naive model and the baseline models as simple as possible, so the fewer the parameters the better.

matchzoo/tasks/classification.py
uduse: I agree, and please be explicit about your point. If you mean `num_classes` shouldn't have a default value, it's a valid point. I thought about that, but since binary classification is the most used classification type, I think it's ok to assume it's the `default`.

matchzoo/tasks/classification.py
uduse: Yes, sure, please be explicit about your point.

matchzoo/tasks/ranking.py
uduse: Yes in this case. Since ultimately we're going to have a  `layers` package in matchzoo, using `keras.layers` is better than using `layers` to avoid confusion.

matchzoo/engine/base_model.py
uduse: I deleted pretty much the whole thing and rewrote it. I agree that docstring is important, but making progress and speed up the review cycle is also important. @bwanglzu we have talked about this and please let the others know the plan. So far I have 100% on the `base_model.py` so that's worth reviewing.

matchzoo/tasks/classification.py
bwanglzu: raise TypeError(what happened?)

matchzoo/tasks/classification.py
bwanglzu: Raise ValueError(explicit)

matchzoo/tasks/ranking.py
bwanglzu: Then why not from keras import layers

matchzoo/engine/base_model.py
bwanglzu: 我只是疑惑这个函数，一样的功能。

matchzoo/models/dense_baseline_model.py
bwanglzu: 所以你觉得是存在用戶不能更改的變量了，在你心中什麼是fixed？我不懂

matchzoo/tasks/ranking.py
uduse: As I said, if you do `from keras import layers`, and use `layers` later, you can't tell if `layers` is come from keras or matchzoo. An alternative is `import keras.layers`, I can do that, but it's not that different.

matchzoo/models/dense_baseline_model.py
uduse: It's just a simple baseline model. The activation functions could be in the params, but I wanted to keep it simple.

matchzoo/tasks/classification.py
uduse: No problem.

uduse: Just a side note, python 3+ prints the full stack trace, so it's ok to have raw exception, since these exceptions always print with the if statements here. But anyways I would like to add those too.

matchzoo/engine/base_model.py
uduse: Well, since I'm rewrote a lot, I wasn't sure that the old docstrings are still accurate. It's better to have no docstrings than misleading docstrings. If it's still accurate once you guys are good with the implementation, I will move back appropriate docstrings you wrote in the old version.

bwanglzu: okay

uduse:
Still missing some docstrings, will leave that for tomorrow.

bwanglzu:
you can try `make test` before sending a commit.

bwanglzu:
also, `coverage` decreased to 94%.

Issue #164, opened 2018-06-09
preprocessor


bwanglzu:
Add pre-processor.py server as a toolkit for text preprocessing.

PR #165, opened 2018-06-09
add pre-processors

bwanglzu:
For issue #164

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,4 +3,4 @@ init:
 
 test:
 	pytest --doctest-modules --doctest-continue-on-failure --cov matchzoo/ --cov-report term-missing --cov-config .coveragerc matchzoo/ tests/
-	flake8 ./matchzoo --exclude __init__.py --ignore=D413,E126
+	flake8 ./matchzoo --exclude __init__.py
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 init:
-		pip install -r requirements.txt --user
+	pip install -r requirements.txt
 
 test:
-		pytest --doctest-modules --doctest-continue-on-failure --cov matchzoo/ --cov-report term-missing --cov-config .coveragerc matchzoo/ tests/
-		flake8 ./matchzoo --exclude __init__.py
+	pytest --doctest-modules --doctest-continue-on-failure --cov matchzoo/ --cov-report term-missing --cov-config .coveragerc matchzoo/ tests/
+	flake8 ./matchzoo --exclude __init__.py --ignore=D413,E126
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -3,6 +3,7 @@
 import re
 import nltk
 import jieba
+import typing
 import many_stop_words
 from functools import reduce
 
@@ -112,7 +113,7 @@ def lemmatization(tokens: list) -> list:
     return [lemmatizer.lemmatize(token, pos='v') for token in tokens]
 
 
-def chain(*funcs):
+def chain(*funcs: typing.Callable) -> typing.Callable:
     """
     Chain a list of functions execute as pipeline.
 
@@ -126,5 +127,6 @@ def chain(*funcs):
     :param *funcs: functions to be executed.
 
     :return rv: return value of the last function.
+
     """
     return lambda x: reduce(lambda f, g: g(f), list(funcs), x)
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -124,7 +124,7 @@ def chain(*funcs):
         ...                   to_lowercase,
         ...                   remove_punctuation,
         ...                   remove_digits)
-        >>> rv = processor("An Example sentence to BE cleaned!)
+        >>> rv = processor("An Example sentence to BE cleaned!")
 
     :param *funcs: list of functions to be executed.
 
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -67,7 +67,7 @@ def remove_punctuation(terms: list) -> list:
     """
     Remove punctuations from list of terms.
 
-    :param terms: list of 
+    :param terms: list of
 
     :return rv: terms  without punctuation.
     """
@@ -92,7 +92,7 @@ def remove_digits(terms: list) -> list:
 
 def stemming(terms: list) -> list:
     """
-    reducing inflected words to their word stem, base or root form.
+    Reducing inflected words to their word stem, base or root form.
 
     :param text: list of string to be stemmed.
     :param mode: stemming algorithm, porter stemer by default.
@@ -124,7 +124,8 @@ def chain(*funcs):
         ...                   to_lowercase,
         ...                   remove_punctuation,
         ...                   remove_digits)
-        >>> rv = processor("This is an Example sentence to BE ! cleaned with digits 31.")
+        >>> rv = processor("This is an Example sentence to /
+        ...                BE ! cleaned with digits 31.")
 
     :param *funcs: list of functions to be executed.
 
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -124,8 +124,7 @@ def chain(*funcs):
         ...                   to_lowercase,
         ...                   remove_punctuation,
         ...                   remove_digits)
-        >>> rv = processor("This is an Example sentence to
-        ...                BE ! cleaned with digits 31.")
+        >>> rv = processor("An Example sentence to BE cleaned!)
 
     :param *funcs: list of functions to be executed.
 
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -124,7 +124,7 @@ def chain(*funcs):
         ...                   to_lowercase,
         ...                   remove_punctuation,
         ...                   remove_digits)
-        >>> rv = processor("This is an Example sentence to /
+        >>> rv = processor("This is an Example sentence to
         ...                BE ! cleaned with digits 31.")
 
     :param *funcs: list of functions to be executed.
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -89,16 +89,25 @@ def remove_digits(tokens: list) -> list:
     return [token for token in tokens if not token.isdigit()]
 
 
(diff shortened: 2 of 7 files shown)


matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+def segmentation(raw: str) -> list:
+    """
+    Chinese sentence segmentation.
+
+    :param raw: raw input given by user.
+
+    :return terms: list of segmentated terms.
+    """
+    return list(jieba.cut(raw))
+
```
uduse: I don't think using two methods for two languages is a good idea. They're both essentially tokenizers. Use an optional language argument and decide what tokenizer backend to use on the fly.

uduse: That's not really doable with the structure you have now since the `reduce` you're using there requires a uniform signature. Either decorate your functions with something like `functools.partial`, or make these function classes and use instances of these classes as pipeline units. I used `reduce` at first, then I switched to decorators, and finally I settled down on classes. Much better flexibility, and easier to serialise.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+
+def to_lowercase(terms: list) -> list:
```
uduse: `terms` should be `tokens`

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    Convert list of terms to lower case.
+
+    :param terms: list of terms.
+
+    :return terms: lower-cased list of terms.
+    """
+    return [term.lower() for term in terms]
+
+
+def remove_stopwords(terms: list, lang: str='en') -> list:
```
uduse: Inconsistent signature, users need to hack with `functools.partial` to use a different language.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Remove stopwords from list of tokenized terms.
+
+    :param terms: list of tokenized terms.
+    :param lang: language code for stopwords.
+
+    :return terms: list of tokenized terms without stopwords.
+    """
+    return [term for term in terms if term not in get_stopwords(lang)]
```
uduse: Maybe using `filter` is more readable.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Remove punctuations from list of terms.
+
+    :param terms: list of
+
+    :return rv: terms  without punctuation.
+    """
+    rv = []
+    for term in terms:
+        term = re.sub(r'[^\w\s]', '', term)
```
uduse: Could compile the regex before using it to speed the whole thing up, but let's leave that for later.

uduse: Could compile this regex before using it to speed the whole thing up, but let's leave that for later.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    return [term for term in terms if not term.isdigit()]
+
+
+def stemming(terms: list) -> list:
+    """
+    Reducing inflected words to their word stem, base or root form.
+
+    :param text: list of string to be stemmed.
+    :param mode: stemming algorithm, porter stemer by default.
```
uduse: Inaccurate docs are confusing. That's why I suggested not having docs for first drafts. I guess you started with multiple arguments but found the necessity of uniform signature later?

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    Chain a list of functions execute as pipeline.
+
+    Examples:
+        >>> processor = chain(tokenizer,
+        ...                   to_lowercase,
+        ...                   remove_punctuation,
+        ...                   remove_digits)
+        >>> rv = processor("An Example sentence to BE cleaned!")
+
+    :param *funcs: list of functions to be executed.
```
uduse: Not really a `list`. It's comma separated values outside and a `tuple` inside.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+
+def remove_punctuation(terms: list) -> list:
+    """
+    Remove punctuations from list of terms.
+
+    :param terms: list of
+
+    :return rv: terms  without punctuation.
+    """
+    rv = []
```
uduse: `rv` means random variable in `numpy`. `ret` is what I usually use and see people use for return value placeholder, `val` is ok, `x` is on par with `rv`, and `cleaned_tokens` (assuming you already changed all `terms` to `tokens`) is the clearest.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+    for term in terms:
+        term = re.sub(r'[^\w\s]', '', term)
+        if term != '':
```
uduse: `if term:` is good enough.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+            rv.append(term)
+    return rv
```
uduse: Maybe the whole thing could be written as something like `list(filter(lambda x: re.sub(r'[^\w\s]', '', x), terms))` but anyways this is clear enough.

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
+"""Matchzoo toolkit for text pre-processing."""
+
+import re
+import nltk
+import jieba
```
uduse: I switch to `thulac` from `jieba`, but that's not a big deal, just keep in mind that maybe multiple backends will be supported in the future.

matchzoo/preprocessors.py
bwanglzu: return value in my case lol

matchzoo/preprocessors.py
bwanglzu: true

matchzoo/preprocessors.py
bwanglzu: only asian langauages such as chinese need segmentation..

bwanglzu: @uduse someting  like return function itself?:
```python```
class Preprocessor...

def func1(self, tokens):
    # do something
    return self
```

matchzoo/preprocessors.py
bwanglzu: done

matchzoo/preprocessors.py
bwanglzu: let's leave it for now..multi language support is not our core features..

matchzoo/preprocessors.py
bwanglzu: done

bwanglzu: done

matchzoo/preprocessors.py
bwanglzu: solved

matchzoo/preprocessors.py
bwanglzu: solved

matchzoo/preprocessors.py
bwanglzu: done

matchzoo/preprocessors.py
bwanglzu: leave it for now

matchzoo/preprocessors.py
```diff
@@ -0,0 +1,133 @@
...
+def remove_punctuation(tokens: list) -> list:
+    """
+    Remove punctuations from list of tokens.
+
+    :param tokens: list of
+
+    :return rv: tokens  without punctuation.
+    """
+    rv = []
+    match = re.compile('[^\w\s]')
```
uduse: Well, the regex still compiles for each call here. Just move the compiled regex outside.

matchzoo/preprocessors.py
uduse: Segmentation and tokenization are really the same thing. They're the processes of partitioning a sequence of characters into minimum processing units (terms). By using class, I mean making callable classes. 

e.g.

```python
class Lower:
    def __call__(self, tokens):
        return [token.lower() for token in tokens]
```

matchzoo/preprocessors.py
bwanglzu: updated

Makefile
```diff
@@ -1,6 +1,6 @@
 init:
-		pip install -r requirements.txt --user
+	pip install -r requirements.txt
 
 test:
-		pytest --doctest-modules --doctest-continue-on-failure --cov matchzoo/ --cov-report term-missing --cov-config .coveragerc matchzoo/ tests/
-		flake8 ./matchzoo --exclude __init__.py
+	pytest --doctest-modules --doctest-continue-on-failure --cov matchzoo/ --cov-report term-missing --cov-config .coveragerc matchzoo/ tests/
+	flake8 ./matchzoo --exclude __init__.py --ignore=D413,E126
```
uduse: So who's making these flake8 standards? I have checks that I don't like as well.

These ignores are better stored in a `setup.cfg` file, so we can comment on what are ignored. See [here](http://flake8.pycqa.org/en/latest/user/configuration.html).

bwanglzu: base model can not pass flake 8 on my laptop

bwanglzu: solved

matchzoo/preprocessors.py
```diff
@@ -35,84 +35,83 @@ def segmentation(raw: str) -> list:
...
 
-    :return terms: list of terms without digits.
+    :return tokens: tokens of tokens without digits.
     """
-    return [term for term in terms if not term.isdigit()]
+    return [token for token in tokens if not token.isdigit()]
 
 
-def stemming(terms: list) -> list:
+def stemming(tokens: list) -> list:
```
faneshion: Is there any possible to make the stemming methods configurable?  For example, if I need to use the 'krovetz' as the stem.

faneshion:
In the function `remove_stopwords`, how about pass the stop_words as the parameters, rather than directly load these words from `get_stopwords`.  As one may have some specific stop word list associated with his task.

matchzoo/preprocessors.py
bwanglzu: done

bwanglzu:
@faneshion `remove_stopwords(tokens=tokens, lang='zh')`

Issue #139, opened 2018-05-29
engine\base_transformer


bwanglzu:
Serve as the base class for data pre-processing.

Need UML design and Python implement.

PR #167, opened 2018-06-11
Feature/base transformer

faneshion:
fix issue #139


matchzoo/inputs/naive_transformer.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if not fixed_vocab:
+            # disable defaultdic behaviour
+            vocabulary = dict(vocabulary)
+            if not vocabulary:
+                raise ValueError("empty vocabulary; perhaps the documents only"
+                                 " contain stop words.")
+        return vocabulary, new_X
+
+    def _limit_vocabulary(self, X, vocabulary, high=None, low=None,
```
bwanglzu: the name and argument of the function is not clear without documentation.
at least, `high` and `low` is strange to use, maybe `lower_boundary` and `upper_boundary` is better?

matchzoo/inputs/naive_transformer.py
```diff
@@ -0,0 +1,169 @@
...
+        """Remove too rare or too common words.
+
+        Prune words that are non zero in more samples than high or less
+        documents than low, modifying the vocabulary, and restricting it to at
+        most the limit most frequent.
+
+        This does not prune samples with zero words.
+        """
+        stop_words = set()
+        if high is None and low is None and limit is None:
```
bwanglzu: `if not high and not low and lot limit` or even better use `any` or `all`

matchzoo/inputs/naive_transformer.py
```diff
@@ -0,0 +1,169 @@
...
+            return X, stop_words
+        return X, stop_words
```
bwanglzu: basically, you create an empty set called `stop_words` and return the empty set to the user, I don't understand.

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+
+            n_original_tokens = len(original_tokens)
+
+            # bind method outside of loop to reduce overhead
+            tokens_append = tokens.append
+            space_join = " ".join
+
+            for n in range(min_n, min(max_n + 1, n_original_tokens + 1)):
+                for i in range(n_original_tokens - n + 1):
+                    tokens_append(space_join(original_tokens[i: i + n]))
```
bwanglzu: I do not prefer this type of `+1`, `-1` `original_tokens[i:i+n]` not pythonic and easy to trigger IndexOutOfRangeError.

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            if not vocabulary:
+                raise ValueError("empty vocabulary passed to fit")
+            self._params['fixed_vocab'] = True
+            self._params['vocabulary'] = dict(vocabulary)
+        else:
+            self._params['fixed_vocab'] = False
+
+    def _check_vocabulary(self):
+        """Check if vocabulary is empty or missing."""
+        if self._params['vocabulary'] is None:
```
bwanglzu: `if not ...` means `if .. is None`

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        """Return the default parameters for transformer."""
+        return engine.TransformerParams()
+
+    def guess_and_fill_missing_params(self):
+        """
+        Guess and fill missing parameters in :attribute:`params`.
+
+        Note: likely to be moved to a higher level API in the future.
+        """
+        if self._params['name'] is None:
```
bwanglzu: `if X` == `if X is None`

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            self._params['name'] = self.__class__.__name__
+
+        if self._params['transformer_class'] is None:
```
bwanglzu: same

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        return self._params
+
+    def _word_ngrams(self, tokens, stop_words=None):
+        """Turn tokens into a sequence of n-grams."""
+        # handle stop words
+        if stop_words is not None:
+            tokens = [w for w in tokens if w not in stop_words]
+
+        # handle token n-grams
+        min_n, max_n = self._params['ngram_range']
```
bwanglzu: `min_n`, `max_n` is not clear

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        if max_n != 1:
+            original_tokens = tokens
```
bwanglzu: `origin_tokens` is a strange variable name

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        else:
+            ngrams = []
+
+        # bind method outside of loop to reduce overhead
+        ngrams_append = ngrams.append
+
+        for n in range(min_n, min(max_n + 1, text_len + 1)):
+            for i in range(text_len - n + 1):
+                ngrams_append(text_document[i: i + n])
+        return ngrams
```
bwanglzu: same, dangerous to use, let's  discuss it later.

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            of words are padded with space.
+        """
+        # normalize white spaces
+        text_document = self._white_spaces.sub(" ", text_document)
+
+        min_n, max_n = self._params['ngram_range']
+        ngrams = []
+
+        # bind method outside of loop to reduce overhead
+        ngrams_append = ngrams.append
```
bwanglzu: I don't understand ...

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+
+        for w in text_document.split():
+            w = ' ' + w + ' '
+            w_len = len(w)
+            for n in range(min_n, max_n + 1):
```
bwanglzu: `n` is `n_gram`? not clear to me

bwanglzu: why `max_n + 1`?

matchzoo/engine/base_transformer.py
uduse: Object boolean checking in params is a special case, where `None` has a special meaning of "not filled yet", and anything else means "already filled". For example, a param like "dropout" could be set to 0, which is reasonable number but its boolean is `False`. I solved this problem locally by making a single param a class, and overrides its `__bool__`, and we can go from there.

tests/unit_test/inputs/test_base_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+from collections import defaultdict
+import copy
+
+
+def test_base_transformer_abstract_instantiation():
+    with pytest.raises(TypeError):
+        transformer = BaseTransformer()
+        assert transformer
+
+def test_base_transformer_concrete_instantiation():
```
uduse: This test is way too complicated. This is usually a sign of test-unfriendly implementation of the target.

uduse: If it's a doc test, which is usually behaviour testing, then I'm perfectly fine with this, just need a little simplification.

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        text_document = self._white_spaces.sub(" ", text_document)
+
+        min_n, max_n = self._params['ngram_range']
+        ngrams = []
+
+        # bind method outside of loop to reduce overhead
+        ngrams_append = ngrams.append
+
+        for w in text_document.split():
+            w = ' ' + w + ' '
```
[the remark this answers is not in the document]
uduse: I think using plus here is better for readability.

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        vocabulary = self._params['vocabulary']
+        if vocabulary is not None:
+            if isinstance(vocabulary, set):
+                vocabulary = sorted(vocabulary)
+            if not isinstance(vocabulary, Mapping):
+                vocab = {}
+                for i, t in enumerate(vocabulary):
+                    if vocab.setdefault(t, i) != i:
+                        msg = "Duplicate term in vocabulary: %r" % t
```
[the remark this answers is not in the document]
uduse: using an intermediate variable here is perfectly fine since the message is long, need to deal with the line width

matchzoo/engine/base_transformer.py
```diff
@@ -0,0 +1,296 @@
...
+                        raise ValueError(msg)
+                vocabulary = vocab
+            else:
+                indices = set(six.itervalues(vocabulary))
+                if len(indices) != len(vocabulary):
+                    raise ValueError("Vocabulary contains repeated indices.")
+                for i in range(len(vocabulary)):
+                    if i not in indices:
+                        msg = ("Vocabulary of size %d doesn't cotain index "
+                               "%d." % (len(vocabulary), i)
```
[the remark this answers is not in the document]
uduse: using `%` is fine. `format` is verbose and hits max line width easily. I personally use python3.6's `f""`, but I guess we don't need a string formatting standard, just use whatever you like.

matchzoo/inputs/naive_transformer.py
```diff
@@ -0,0 +1,169 @@
...
+            # id1 = instance['id1']
+            # id2 = instance['id2']
+            pair_feature = ([], [])
+            for tid, text in enumerate(pair):
+                for feature in analyze(text):
+                    try:
+                        feature_idx = vocabulary[feature]
+                        pair_feature[tid].append(feature_idx)
+                    except KeyError:
+                        # Ignore out-of-vocabulary items for fixed_vocab=True
```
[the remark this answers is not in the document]
uduse: Agree. At least prints it.

matchzoo/inputs/naive_transformer.py
uduse: `high`, `low` are fine when you see "limit" in the method's name. A bigger problem is the order of these two things, where naturally `low` should on the left side and `high` on the other side. Otherwise, if these two arguments are passed positionally instead of as keywords, it would be confusing.

uduse: wait, so `high` and `low` are term frequencies? Now it's confusing...

bwanglzu:
@faneshion Shall  we close this one and re-design the fit_transform according to what we've discussed yesterday?

matchzoo/inputs/naive_transformer.py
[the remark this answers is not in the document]
bwanglzu: @uduse Shall we use `logiger` or `warning` library?

PR #169, opened 2018-06-12
Update on parameters and hyper parameter tuner.

uduse:
For issue #170 

- feature: parameter validation
- feature: parameter hyper space setting
- feature: auto hyper parameter tuning
- fix: 'model_class' parameter auto setting


matchzoo/engine/base_model.py
```diff
@@ -45,22 +48,34 @@ def get_default_params(cls) -> engine.ModelParams:
...
         """
-        return engine.ModelParams()
+        params = engine.ParamTable()
+        params.add(engine.Param('name'))
+        params.add(engine.Param('model_class', cls))
+        params.add(engine.Param('input_shapes'))
+        params.add(engine.Param('task'))
+        params.add(engine.Param('metrics'))
+        params.add(engine.Param('loss'))
+        params.add(engine.Param('optimizer'))
```
bwanglzu: I'm confused by this part, we add Param class instance to param table here without a value (only key)

matchzoo/engine/base_model.py
```diff
@@ -98,7 +113,8 @@ def fit(
             x: typing.Union[np.ndarray, typing.List[np.ndarray]],
             y: np.ndarray,
             batch_size: int = 128,
-            epochs: int = 1
+            epochs: int = 1,
```
bwanglzu: I guess we should configure default `epoch`  as a larger value? something like 20 as default, if user forget to set `epochs` when fitting a model, at least it will not end-up with 1 epoch.

matchzoo/engine/tuner.py
```diff
@@ -0,0 +1,54 @@
...
+    def tune(self):
+
+        def test_wrapper(space):
+            for key, value in space.items():
+                self._model.params[key] = value
+            self._model.guess_and_fill_missing_params()
+            self._model.build()
+            # TODO: use model fit loss instead of a random loss
+            return {'loss'  : random.random(), 'space': space,
+                    'status': hyperopt.STATUS_OK}
```
bwanglzu: we're going to remove the `test_wrapper`, right?

matchzoo/engine/param_table.py
```diff
@@ -0,0 +1,55 @@
...
+            param.name: param.hyper_space
+            for param in self._params.values()
+            if param.hyper_space is not None
+        }
+
+    def __getitem__(self, key: str) -> typing.Any:
+        return self._params[key].value
+
+    def __setitem__(self, key: str, value: typing.Any):
+        self._params[key].value = value
```
bwanglzu: why:
```python
self._params[key].value ? # instead of self._param[key]
```
It's a nested dict?

matchzoo/engine/base_model.py
uduse: Yes, only key. The value is `None` by default, which means "not filled."

matchzoo/engine/base_model.py
uduse: `epoch` defaults to 1 is better, since not setting it means "I am not thinking about multiple epochs." Also, `keras` defaults to 1, and I kept it consistent (also for `batch_size`).

matchzoo/engine/tuner.py
uduse: There must be a function to interface `fmin`, but will eventually look very different to what I have now. This version is just a simple demo. It works.

matchzoo/engine/base_model.py
bwanglzu: okay

matchzoo/engine/param_table.py
uduse: - `self._params` is a `ParamTable`
- `self._params[key]` is a `Param` instance
- `self._params[key].value` accesses the value in the `Param` instance

also, it's important to trigger `Param.value.setter` since validation happens there.

uduse:
@bwanglzu Validators are mostly `lambda`s, and `dill` serializes `lambda`s while `pickle` does not. This branch was created before the feature request issue, and was originally for some docstring updates, thus the name.

matchzoo/engine/param_table.py
bwanglzu: make sense to me

uduse:
`Tuner` is perfectly fine to be a standalone function instead of a class, so I switched.

matchzoo/engine/hyper_spaces.py
```diff
@@ -0,0 +1,109 @@
...
+
+        :param options: options to search from
+        """
+        super().__init__(hyperopt_func=hyperopt.hp.choice, options=options)
+
+
+class quniform(HyperoptProxy):
+    """:function:`hyperopt.hp.quniform` proxy."""
+
+    def __init__(self, low, high, q=1):
```
bwanglzu: type hints

uduse:
`int` and `float` both inherits `numbers.Number`. Third party library data types like `numpy.int64` is also a subclass of `numbers.Number`. All of them should work in a parameter hinted as type `numbers.Number`.

faneshion:
Overall, I like the hyper-parameter tuner. For all model specific hyper-parameters, it will require the model define the default value space.

matchzoo/engine/base_model.py
```diff
@@ -30,7 +33,7 @@ def __init__(self, params: typing.Optional[engine.ModelParams] = None,
         self._backend = backend
 
     @classmethod
-    def get_default_params(cls) -> engine.ModelParams:
+    def get_default_params(cls) -> engine.ParamTable:
```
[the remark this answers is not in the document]
uduse: Since `Param` has `__bool__`, you can do `if param` to check it is set or not.

Issue #138, opened 2018-05-29
models\dssm


bwanglzu:
Create a sample model to use MatchZoo 2.0 base_model.

bwanglzu:
Start to work on this issue until base_model settle down.

PR #186, opened 2018-06-20
add dssm model

bwanglzu:
for issue #138 

Adapt dssm to new `base_model`.


matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params()
+        params.add(engine.Param('w_initializer', 'glorot_normal'))
+        params.add(engine.Param('b_initializer', 'zeros'))
+        # TODO GET TRI-LETTER DIMENSIONALITY FROM FIT-TRANSFORM AS INPUT SHAPE
+        params.add(engine.Param('input_shapes', [(30000,), (30000,)]))
```
uduse: Should do `params['input_shapes'] = ...` instead. I'll add a checking for already added params in a PR later.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('dim_fan_out', 128))
+        params.add(engine.Param('dim_hidden', 300))
+        params.add(engine.Param('activation_hidden', 'tanh'))
+        params.add(engine.Param('activation_prediction', 'softmax'))
+        params.add(engine.Param('lr', 0.01))
```
uduse: I am thinking about relaxing the constraint of "using only native data types or matchzoo data types" for params. This way, learning rate is no longer a param itself. It is and it should be one of the parameters of `params['optimizer']`. So when the user really wants to tweaking learning rate, the user can do `params['optimizer'] = keras.optimizers.SGD(learning_rate=0.01)` or something. It's not friendly to parameter tuning though, but I'll figure that out separately.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('decay', 1e-6))
```
uduse: Same as `lr`.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('batch_size', 1024))
+        params.add(engine.Param('num_epochs', 20))
```
uduse: `batch_size` and `num_epochs` are not part of model structures. We should set them manually for now and add them into a higher level API to models (e.g. something called a `Trainer`) later.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('train_test_split', 0.2))
+        params.add(engine.Param('verbose', 1))
```
uduse: `verbose`? Not needed for now. We can do all the loggers later and make `verbose` a shared parameter in the base model.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('shuffle', True))
```
uduse: Also, not a part of the model structure.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('optimizer', 'sgd'))
+        params.add(engine.Param('num_hidden_layers', 2))
+        params.add(engine.Param('loss', 'categorical_crossentropy'))
```
uduse: It's based on `params['task']`.

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
...
+        params.add(engine.Param('activation_hidden', 'tanh'))
+        params.add(engine.Param('activation_prediction', 'softmax'))
+        params.add(engine.Param('lr', 0.01))
+        params.add(engine.Param('decay', 1e-6))
+        params.add(engine.Param('batch_size', 1024))
+        params.add(engine.Param('num_epochs', 20))
+        params.add(engine.Param('train_test_split', 0.2))
+        params.add(engine.Param('verbose', 1))
+        params.add(engine.Param('shuffle', True))
+        params.add(engine.Param('optimizer', 'sgd'))
```
uduse: Use `params['optimizer'] = 'sgd'`.

tests/unit_test/models/test_dssm_model.py
```diff
@@ -0,0 +1,44 @@
...
+    assert model.predict(x) is not None
+
+
+def test_save_load_model(model):
+    tmpdir = '.tmpdir'
+    model.save(tmpdir)
+    assert engine.load_model(tmpdir)
+    with pytest.raises(FileExistsError):
+        model.save(tmpdir)
+    shutil.rmtree(tmpdir)
```
uduse: I don't think each model needs its own tests. A simple doctest should give 100% coverage already.

matchzoo/models/dssm_model.py
bwanglzu: when to use `params setter`, and when to use `engine.Param`?

matchzoo/models/dssm_model.py
bwanglzu: that's better

matchzoo/models/dssm_model.py
bwanglzu: update I get the idea

matchzoo/models/dssm_model.py
```diff
@@ -0,0 +1,88 @@
+"""An implementation of DSSM, Deep Structured Semantic Model."""
+from matchzoo import engine
+from keras.models import Model
+from keras.layers import Dense, Input, Dot
+
+
+class DssmModel(engine.BaseModel):
```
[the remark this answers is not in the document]
bwanglzu: done

matchzoo/models/dssm_model.py
bwanglzu: ok let's discuss later

matchzoo/models/dssm_model.py
bwanglzu: done

tests/unit_test/models/test_dssm_model.py
bwanglzu: if we remove this, the coverage check is going to fail

matchzoo/models/dssm_model.py
bwanglzu: done

matchzoo/models/dssm_model.py
bwanglzu: do something like this?
```python
params['task'] = 'classification'
params['loss']='categorical_crossentropy'
```

bwanglzu: that's correct

tests/unit_test/models/test_dssm_model.py
uduse: #191 I just added standardized model testing procedure in this PR. Testing a new model with one line.

bwanglzu: okay @uduse wait unitl #191  merged.

matchzoo/models/dssm_model.py
```diff
@@ -39,44 +38,51 @@ def get_default_params(cls) -> engine.ParamTable:
...
+        x = [base_network(input_left),
+             base_network(input_right)]
         # Dot product with cosine similarity.
-        x = Dot(axes=1,
-                normalize=True)(x_in)
-        x_out = Activation(activation=self._params['activation_prediction'])(x)
-        self._backend = Model(inputs=x_in, outputs=x_out)
+        x = Dot(axes=[1, 1],
+                normalize=True)(x)
+        x_out = Dense(2,
```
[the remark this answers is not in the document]
bwanglzu: @faneshion Then for the ranking case, we need to sue `sigmoid` instead of `softmax` to transform the result to prob, or write our activation.

matchzoo/models/dssm_model.py
```diff
@@ -39,44 +38,51 @@ def get_default_params(cls) -> engine.ParamTable:
...
 
-        :return: 128 dimension vector representation.
+        :return: x: 128d vector(tensor) representation.
         """
-        x_in = Input(shape=(dim_triletter,))
-        # Initialize input layer.
-        x = Dense(units=self._params['dim_hidden'],
-                  input_shape=(dim_triletter,),
+        input = Input(shape=input_shape)
+        x = Dense(self._params['dim_hidden'],
```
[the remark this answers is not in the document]
bwanglzu: interesting, let's discuss it. might affect varies models.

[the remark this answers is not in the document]
bwanglzu: @faneshion I've checked the keras API, the `sparse` argument is not available at the moment (it's there, but might cause bugs). Let's use normal Input layer at the moment, and refactor in the future.

Issue #187, opened 2018-06-21
Undefined Var


wqh17101:
in MatchZoo/matchzoo/inputs/point_generator.py line 290，there is a undefined var 'randomly' which cause NameError: global name 'randomly' is not defined

wqh17101:
And i set it False,it can work now.
i try to run anmm.it seems that everything is ok at the beginning.
![image](https://user-images.githubusercontent.com/26429138/41696752-4ceee050-7548-11e8-9eb1-47d15c1758e1.png)
But some epochs latter,it didnt work.
![image](https://user-images.githubusercontent.com/26429138/41696792-7d63b148-7548-11e8-9523-23ae1cf41aa0.png)

wqh17101:
every model which uses DRMM_PointGenerator has this problem.

faneshion:
It's really a bug in data generation. Thanks for your help.  The `DRMM_PointGenerator` is used under classification task (i.e., QuoraQP) by DRMM and ANMM. I will fixed it right now.

wqh17101:
I will try.

PR #189, opened 2018-06-21
fixed DRMM_PointGenerator randomly

faneshion:
#187


bwanglzu:
@faneshion Correct me if I'm wrong, `is_train` is a `str` instead of `bool`.
And:
```python
if self.is_train
```
will always return `True`.   They why do we need `if`?

faneshion:
uhhh. Right now, `DRMM_PointGenerator` is used for classification, however, it can also be used for ranking. `is_train` is a bool which is defined as `self.is_train = config['phase'] == 'TRAIN'`.

PR #191, opened 2018-06-22
Feature/misc updates

uduse:
feature: set evaluation verbosity
feature: set task output shape
feature: repetitive parameter name checking
feature: standardize model testing
refactor: remove outdated test
refactor: move `all_params_filled` to `ParamTable` as a method


matchzoo/engine/param_table.py
```diff
@@ -64,3 +73,20 @@ def __str__(self):
     def __iter__(self) -> typing.Iterator:
         """:return: A iterator that iterates over all parameter instances."""
         yield from self._params.values()
+
+    def completed(self) -> bool:
```
bwanglzu: `completed` is a little bit strange for me, maybe `filled` is a better name?

uduse: I though about that, but chose `model.params.completed` because it's less ambiguous. `model.params.filled` doesn't necessarily convey the idea of **all** parameters being filled, but `model.params.completed` does. Same reason why you can't directly convert a numpy array into a bool, but have to use `numpy.ndarray.any` or `numpy.ndarray.all`.

bwanglzu: Not a big issue

uduse: I chose names very carefully, so when you see a name change, there's definitely some consideration behind it.

uduse:
@bwanglzu It's a problem needs to be addressed. I wanted to decouple tasks and keras layers, but also wanted to keep it simple. That's why the awkward `make_output_layer` method exists. Since there's solid use case where same task makes different output layers in different models, it's time to refactor.

bwanglzu:
@uduse no problem,

Issue #175, opened 2018-06-13
Design DataPack Interface


pl8787:
![image](https://user-images.githubusercontent.com/1762835/41358774-ae6eb9ce-6f5b-11e8-9d81-0790dfc89e1a.png)

bwanglzu:
The idea is to use `ProcessUnit` as atom element to handle pre-process actions, and **Chain** a  sequence of units based on requirements. Some pseudo code:
```python
class ProcessUnit(object):
    @staticmethod
    @abc.abstractmethod
    def _process(input):
        return

    @abc.abstractmethod
    def fit(self, input):
        return

    @abc.abstractmethod
    def transform(self, input):
        return

    @classmethod
    def fit_transform(cls, input):
        ctx = cls.fit(input)
        rv = cls.transform(input)
        return rv, ctx # rv is return value of transform action, ctx is statistics of fit action.
```

bwanglzu:
To create an `ProcessUnit`, inherit from `ProcessUnit` class. For example, to **tokenize** user input, we create a `TokenizeUnit` class:

```python
class TokenizeUnit(ProcessUnit):

    def _process(input):
        return nltk.word_tokenize(input)

    def fit(self, input):
        return # fit is not needed for tokenization

    def transform(self, input):
        return TokenizeUnit._process(input)
```

bwanglzu:
For user input, we specify a sequence (ordered) of `ProcessUnit`, and add these Units into a `Chian` class:

```python
class Chain(collections.deque):
    """The class chain an ordered sequence of ProcessUnit"""
    pass

processors = Chain(
    TokenizeUnit(),
    PuncRemovalUnit() # Remove punctuation from input.
    DigitRemovalUnit() # Remove digits from input.
    StemmingUnit # Stemming.
    ...
)   
```

bwanglzu:
@aneesh-joshi @thiziri How do you think the idea for MZ 2.0 data pre-processing? (through it will not exposed to users)

uduse:
Just dumping some ideas here:

- `processor = Chain(A, B, C)` is weird. Try something like `processor = Processor(pipeline=[A, B, C])`
- Processing units may have the following structure: 
    ```
                 ProcessingUnit
                   (+ abstract method transform)
                  /                  \
    StatelessUnit            StatefulUnit
                                         (+ attribute ctx)
                                         (+ abstract method fit)
    ``` 
[...]

PR #192, opened 2018-06-22
First design of datapack

bwanglzu:
for #175 

First design of DataPack, allow list of pairwise-input and context as parameters, inherit from pandas.DataFrame.


matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+                 columns: list=['text_left', 'text_right'],
+                 dtype: object=None,
+                 copy: bool=True):
+        """Initialize."""
+        super(DataPack, self).__init__(data=data,
+                                       index=index,
+                                       columns=columns,
+                                       dtype=dtype,
+                                       copy=copy)
+        if self.shape[1] != 2:
```
pl8787: This constrain means that DataPack can process two texts data, however if we have additional features of text pairs, such as TF-IDF, BM25 ... 

Maybe remove this constrain?

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        >>> type(dp.sample(1))
+        <class 'matchzoo.datapack.DataPack'>
+        >>> dp.size
+        2
+        >>> features, context = dp.unpack()
+    """
+
+    _metadata = ['context']
+
+    DATA_FILENAME = 'data.pkl'
```
uduse: Just checked my `base_model.py` and it has the same problem: when you save with `dill`, the filename should be something like `data.dill`.

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        {'vocab_size': 2000}
+        >>> # sample without replacement for generation.
+        >>> type(dp.sample(1))
+        <class 'matchzoo.datapack.DataPack'>
+        >>> dp.size
+        2
+        >>> features, context = dp.unpack()
+    """
+
+    _metadata = ['context']
```
uduse: What for?

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
+"""Matchzoo DataPack, pair-wise tuple (feature) and context as input."""
+
+import typing
+from pathlib import Path
+
+import dill
+import pandas as pd
+
+
+class DataPack(pd.DataFrame):
```
uduse: Using `pandas.DataFrame` is great but inheritance is a good idea here. Use delegation instead.

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+                                       columns=columns,
+                                       dtype=dtype,
+                                       copy=copy)
+        if self.shape[1] != 2:
+            raise ValueError(
+                "Pair-wise input expected.")
+        self.context = context
+
+    @property
+    def _constructor(self) -> callable:
```
uduse: What for?

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        """Subclass pd.DataFrame."""
+        return DataPack._internal_ctor
+
+    @classmethod
+    def _internal_ctor(cls, *args, **kwargs):
```
uduse: These two constructors are really confusing.

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        """Create subclass inputs to store context."""
+        kwargs['context'] = None
+        return cls(*args, **kwargs)
+
+    @property
+    def size(self) -> int:
```
uduse: Override `__len__` instead.

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        """Get size of the data pack."""
+        return self.shape[0]
+
+    def unpack(self) -> typing.Union[pd.DataFrame, dict]:
```
uduse: Redundant method. When you have `self`, you also have `self.context`.

matchzoo/datapack.py
```diff
@@ -0,0 +1,114 @@
...
+        if dirpath.exists():
+            raise FileExistsError
+        else:
+            dirpath.mkdir()
+
+        data_file_path = dirpath.joinpath(self.DATA_FILENAME)
+        dill.dump(self, open(data_file_path, mode='wb'))
+
+        context_file_path = dirpath.joinpath(self.CONTEXT_FILENAME)
+        dill.dump(self.context, open(context_file_path, mode='wb'))
```
uduse: When you save `self`, `self.context` is also saved, so there's no need to save two files.

matchzoo/datapack.py
bwanglzu: @pl8787 got it.

matchzoo/datapack.py
bwanglzu: pandas set the restrictions see: http://pandas.pydata.org/pandas-docs/stable/extending.html#extending-subclassing-pandas

matchzoo/datapack.py
bwanglzu: got it

matchzoo/datapack.py
bwanglzu: define a proxy class?

matchzoo/datapack.py
bwanglzu: same as `_metadata`.

matchzoo/datapack.py
bwanglzu: same as `_metadata`.

matchzoo/datapack.py
bwanglzu: yes i thought about it, actually `len(DataPack)` is strange from my side, since the term "DataPack" sounds like a multi-dimensional dataframe instead of an array or list. I believe size is better.

matchzoo/datapack.py
bwanglzu: their is some redundancy here. taken into consider the user might not aware the `context` variable, I believe `unpack` is more natural method to access the data.

To some extent, i think sacrifice is worthwhile.

matchzoo/datapack.py
bwanglzu: good one

matchzoo/datapack.py
uduse: @bwanglzu Kinda of. It's more than a proxy class because it has its own special behaviors.

matchzoo/datapack.py
uduse: Ok so that makes inheritance here even less attractive.

matchzoo/datapack.py
uduse: Using `size` doesn't solve that problem, since your size always returns a single `int`, it doesn't fit the multi-dimensional nature of the structure very well too. Also, `__len__` is used by `__bool__`, so you also can do `if data_pack: ...` this way. However, if you switch to delegation and define a property that exposes the pandas dataframe backend, then you can do `data_pack.df.size` when you need it, but it has different behavior since it multiplies two dimensions, and it means "number of elements" while what you intended to express is "number of samples". Anyways, I think you should take out the inheritance first.

matchzoo/datapack.py
uduse: I don't think so. Letting users know there's a `context` property takes same effort as letting users know there's a `unpack` method. And using `unpack` doesn't even save a single line of code: `features, context = dp.unpack()` is exactly the same as `features, context = dp, dp.context`. It even uses one more line since when the user know there's a `context` property, the user won't need to unpack or something, just directly access it when needed. Plus the type hint is wrong, it should be a `tuple` type. Plus the type hint is a bit confusing: yes, you're returning a `pd.DataFrame` since it's a subclass, but `DataPack` servers very different purposes and should be mentioned explicitly.

matchzoo/datapack.py
```diff
@@ -0,0 +1,104 @@
...
+    :param dirpath: directory path of the saved model
+    :return: a :class:`DataPack` instance
+    """
+    dirpath = Path(dirpath)
+
+    data_file_path = dirpath.joinpath(DataPack.DATA_FILENAME)
+    dp = dill.load(open(data_file_path, 'rb'))
+    data, context = dp.unpack()
+
+    return DataPack(data=data, context=context)
```
uduse: I doubt the saving and loading here are correct reverse functions of each other. Let's see some tests.

bwanglzu: it works, I'll add tests.

matchzoo/datapack.py
```diff
@@ -0,0 +1,116 @@
...
+        :param new_data_pack: A new DataPack object.
+        """
+        new_dataframe = new_data_pack.dataframe
+        new_context = new_data_pack.context
+        self._dataframe = self._dataframe.append(
+            new_dataframe,
+            ignore_index=True)
+        self._context.update(new_context)
+
+    def save(self, dirpath: typing.Union[str, Path]):
```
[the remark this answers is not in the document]
bwanglzu: When you save self, self.context is also saved, so there's no need to save two files.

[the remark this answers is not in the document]
bwanglzu: When you save self, self.context is also saved, so there's no need to save two files.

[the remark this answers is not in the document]
bwanglzu: Ok

matchzoo/datapack.py
```diff
@@ -0,0 +1,116 @@
...
+        Sample records from `DataPack` object, for generator.
+
+        :param number: number of records to be sampled, use `batch_size`.
+        :param replace: sample with replacement, default value is `True`.
+
+        :return data_pack: return `DataPack` object including sampled data
+                           and context.
+        """
+        return DataPack(self._dataframe.sample(n=number, replace=replace),
+                        self._context)
```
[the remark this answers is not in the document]
bwanglzu: Don’t understand

matchzoo/datapack.py
```diff
@@ -0,0 +1,116 @@
...
+
+    :param dirpath: directory path of the saved model
+    :return: a :class:`DataPack` instance
+    """
+    dirpath = Path(dirpath)
+
+    data_file_path = dirpath.joinpath(DataPack.DATA_FILENAME)
+    dp = dill.load(open(data_file_path, 'rb'))
+
+    return DataPack(data=dp.dataframe, context=dp.context)
```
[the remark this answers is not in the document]
bwanglzu: Will change

tests/unit_test/test_datapack.py
```diff
@@ -0,0 +1,30 @@
...
+import shutil
+
+@pytest.fixture
+def data_pack():
+    data = [([1,3], [2,3]), ([3,0], [1,6])]
+    ctx = {'vocab_size': 2000}
+    return DataPack(data, ctx)
+
+def test_length(data_pack):
+    assert len(data_pack) == 2
```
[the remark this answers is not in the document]
bwanglzu: Okay

matchzoo/datapack.py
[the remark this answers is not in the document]
bwanglzu: added an allow_overwrite argument.

matchzoo/datapack.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        new_dataframe = new_data_pack.dataframe
+        new_context = new_data_pack.context
+        self._dataframe = self._dataframe.append(
+            new_dataframe,
+            ignore_index=True)
+        if overwite_context:
+            self._context = {**self._context, **new_context}
+        else:
+            self._context = {**new_context, **self._context}
```
uduse: This is weird. (1) why not use `dict.update` (2) `overwrite_context` seems redundant, since you can achieve the same thing by switching two data packs.

matchzoo/datapack.py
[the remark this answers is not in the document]
uduse: Can't find any change here.

matchzoo/datapack.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self,
+                 data: list,
+                 context: dict={}):
+        """Initialize."""
+        self._dataframe = pd.DataFrame(data)
+        self._context = context
+
+    def __len__(self) -> int:
+        """Get size of the data pack."""
```
uduse: "size" means number of elements in pandas, here you mean the number of samples, or the number of rows.

matchzoo/datapack.py
```diff
@@ -0,0 +1,115 @@
...
+                        self._context.copy())
+
+    def append(self, new_data_pack: 'DataPack'):
+        """
+        Append a new `DataPack` object to current `DataPack` object.
+
+        It should be noted that the context of the previous `DataPack`
+        will be updated by the new one.
+
+        :param new_data_pack: A new DataPack object.
```
uduse: It's not really a "new" `DataPack`, which implies that the input data pack will **replace** the current data pack, for example, in property setters, those are "new" values. I guess what you intended to express is "The DataPack to append"

matchzoo/datapack.py
```diff
@@ -0,0 +1,115 @@
...
+        :param overwite_context: Allow overwrite common context by new context.
+        """
+        new_dataframe = new_data_pack.dataframe
+        new_context = new_data_pack.context
+        self._dataframe = self._dataframe.append(
```
uduse: Same issue. `self._dataframe.append(new_dataframe, ignore_index=True)` this is the "new_dataframe" since it's the data frame that overwrites the current one.

matchzoo/datapack.py
```diff
@@ -0,0 +1,115 @@
...
+            new_dataframe,
+            ignore_index=True)
+        self.context.update(new_context)
+
+    def save(self, dirpath: typing.Union[str, Path]):
+        """
+        Save the `DataPack` object.
+
+        A saved `DataPack` is represented as a directory with two files.
```
uduse: Out dated docs.

PR #193, opened 2018-06-23
Refactor output layer

uduse:
Refactor `tasks.BaseTask.make_output_layer` into `engine.BaseModel._make_output_layer`, so models can customize their output layers based on different task by overriding the default behavior.

### DIFF
--- a/matchzoo/engine/base_model.py
+++ b/matchzoo/engine/base_model.py
@@ -9,6 +9,7 @@ import numpy as np
 import keras
 
 from matchzoo import engine
+from matchzoo import tasks
 
 
 class BaseModel(abc.ABC):
@@ -226,6 +227,14 @@ class BaseModel(abc.ABC):
 
         if self._params['optimizer'] is None:
             self._params['optimizer'] = 'adam'
+
+    def _make_output_layer(self):
+        """:return: a correctly shaped keras dense layer for model output."""
+        task = self._params['task']
+        if isinstance(task, tasks.Classification):
+            return keras.layers.Dense(task.num_classes, activation='softmax')
+        elif isinstance(task, tasks.Ranking):
+            return keras.layers.Dense(1, activation='sigmoid')
 
 
 def load_model(dirpath: typing.Union[str, Path]) -> BaseModel:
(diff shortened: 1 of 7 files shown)


matchzoo/engine/base_model.py
```diff
@@ -235,6 +235,8 @@ def _make_output_layer(self):
             return keras.layers.Dense(task.num_classes, activation='softmax')
         elif isinstance(task, tasks.Ranking):
             return keras.layers.Dense(1, activation='sigmoid')
```
faneshion: It's wonderful to bind the output with the task type. Here, I think the `sigmoid` is optimal in `tasks.Ranking`, since not all models require the output to be a probability. Sometimes, it is more efficient to train the model when the output to be a real number (e.g., in hinge loss).

uduse: The point of moving layer constructions from tasks to models is to allow models to use different losses for the same task by overriding the default `_make_output_layer`. In your case where a model wants a different output activation function, the model will need to override `_make_output_layer` and handle `tasks.Ranking` output layer construction itself.

faneshion: Well, it's well enough in current `_make_output_layer` since it can cover the majority of ranking cases. Special requirement can be satisfied by overriding.

matchzoo/models/dense_baseline_model.py
```diff
@@ -38,5 +38,5 @@ def build(self):
         x = keras.layers.concatenate(x_in)
         x = keras.layers.Dense(
                 self._params['num_dense_units'], activation='relu')(x)
-        x_out = self.params['task'].make_output_layer()(x)
+        x_out = self._make_output_layer()(x)
```
bwanglzu: why use private method instead of a public one?

matchzoo/models/naive_model.py
```diff
@@ -13,5 +13,5 @@ def build(self):
         x_in = [keras.layers.Input(shape)
                 for shape in self._params['input_shapes']]
         x = keras.layers.concatenate(x_in)
-        x_out = self._params['task'].make_output_layer()(x)
+        x_out = self._make_output_layer()(x)
```
bwanglzu: same

matchzoo/engine/base_model.py
uduse: @faneshion correct, that's the whole point of this PR.

Issue #176, opened 2018-06-13
Caregorize ProcessUnit


bwanglzu:
If everyone agree the way we design for our first version of `pre-processor` (see #175 ), we need to design the (potentially nedded) `ProcessUnits`.

Roughly, we have three type of `ProcessUnit`:

1. The input data is **raw** text, the `ProcesUnit` should be used for segmentation/tokenization. 

2. `ProcessUnit` designed for text normalization, such as stemming, stop words removal, white space removal etc..

3. `ProcessUnit` designed for post-normalization activities, such as build vocabulary (need discuss)

wqh17101:
We need to provide a Unit which can filter some words from users' customized word list for them

uduse:
@wqh17101 a unit that filter words is easy enough to write your own, so I don't think it has to be in matchzoo.

PR #199, opened 2018-06-25
refactor stateful unit, update nletterprocessunit, update unittest

bwanglzu:
For feature #176 
1. Add `NgramLetterProcessorUnit` to `Processors` to generate tri-letters (dependent on `N`).
2. Add unit tests.
3. Refactor unit test for processors.

### DIFF
--- a/matchzoo/preprocessors.py
+++ b/matchzoo/preprocessors.py
@@ -28,10 +28,12 @@ class StatefulProcessorUnit(ProcessorUnit, metaclass=abc.ABCMeta):
     @property
     def state(self):
         """Get current state."""
+        if not self._state:
+            self.fit()
         return self._state
 
     @abc.abstractmethod
-    def fit(self, input: typing.Any) -> dict:
+    def fit(self) -> dict:
         """Abstract base method, need to be implemented in subclass."""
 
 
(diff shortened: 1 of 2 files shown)


matchzoo/preprocessors.py
```diff
@@ -195,6 +195,8 @@ def transform(self, tokens: list, ngram: int=3) -> list:
 
         :param tokens: list of tokens to be transformed.
         :param ngram: By default use 3-gram (tri-letter).
+
+        :return: set of tri-letters, dependent on `ngram`.
```
faneshion: In the current preprocessors, all operations are conducted on raw texts. However, we have to map each word/n-letters to id, which is  the function of `Vocabulary`.  Thus, we need to design a `Vocabulary` class, whose `fit` construct a _word_dict and `transform` map list of string to list of ids.

uduse: @faneshion fire up an issue for that?

uduse:
Just dumping more rationale here: `fit` and `transform` should focus on their on jobs. `fit` alters processor unit's inner state without changing input, while `transform` returns a copy of the transformed input without changing inner state. This might be inefficient for some processor units (e.g. n-grams), but it's more important to draw the line between these two things, and performance is not our concern for now. Nonetheless, that inefficiency can be reduced by inducing something like a `fit_transform` in the future, but we don't need that for now.

matchzoo/preprocessors.py
faneshion: OK

bwanglzu:
logically, `fit` should happens before `transform`. While in reality, this doesn't make sense.

If you want to develop a `TrigramProcessUnit` to create `tri-letters` (word -> wor, ord ..), the context (state) you wanna fit is the length of the transformed `tri-letters`. This wouldn't happen before `transform` has been called.

Say if you force to separate `fit` and `transform` operation, then you have two options:
1. Create `tri-letters` again in fit, get it's length of `tri-letters` and throw away the `tri-letters`, computationally expensive.
2. Move `fit` into the follow-up `Process Unit`, then we lose the independence of each units.

uduse:
(1) is exactly what I meant. We do it twice. Quote:

> This might be inefficient for some processor units (e.g. n-grams), but it's more important to draw the line between these two things, and performance is not our concern for now. Nonetheless, that inefficiency can be reduced by inducing something like a fit_transform in the future, but we don't need that for now.

bwanglzu:
This will be inefficient for every `StatefulProcessorUnit`..

uduse:
counter example: [the vocabulary class I just wrote](https://github.com/faneshion/MatchZoo/blob/feature/vocabulary/matchzoo/preprocessors.py#L171)

uduse:
One more thing. The `input` of a `fit` might be different from the `input` of a `transform`, or a number of `transform`s. A common usage is a single `fit` on training data followed by a `transform` on both training and testing data.

matchzoo/preprocessors.py
```diff
@@ -166,3 +166,65 @@ def transform(self, tokens: list) -> list:
...
+    Process unit for n-letter generation.
+
+    Triletter is used in :DSSMModel: and :CDSSMModel:.
+    This processor is expected to execute after `Vocab`
+    has been created.
+
+    Returned `input_dim` is the dimensionality of :DSSMModel:.
+    """
+
+    def __init__(self):
```
uduse: Not needed.

PR #200, opened 2018-06-26
override  method for some layer. fix some bug about optimizers

stanpcf:
two changes:
1. override `get_config` method for some layers, which layers cannot be loaded from model-weights if don't have those config.
2. fix bug about optimizers in mz.


bwanglzu:
@pl8787 can you review?

Issue #206, opened 2018-06-27
Vocabulary class for preprocess


faneshion:
Here, we need to design a `Vocabulary` class in the preprocess. The `Vocabulary` class is to map a list of words into a list of word-ids. Maybe there can be a 'fit' function to store '_word_dict', and a 'transform' function to map words to ids. We can discuss it more~ @uduse @pl8787 @yangliuy @bwanglzu

bwanglzu:
Elaborate a little bit more on `word_dict`?

uduse:
`StatefulProcessorUnit` suffices. A quick draft here: https://github.com/faneshion/MatchZoo/blob/feature/vocabulary/matchzoo/preprocessors.py#L171

faneshion:
The __missing__ function is a good idea to deal with OOV words. Can the `fit` function increasingly build the `token_index` and `index_token`? In most times, we would call the `fit` several times to build the final `vocabulary`.

uduse:
@faneshion Yes that's a common usage. However, a `fit` usually resets the processor's inner state. 

e.g. sklearn's `LabelEncoder`.
![image](https://user-images.githubusercontent.com/4717005/42071225-de14aa02-7b8c-11e8-910d-c189a2fd078b.png)

It's important to keep the interface consistent. Either we don't reset inner state for all stateful processors, or we implements such incremental fitting for all of them. The latter one is harder since one needs to keep track of more things. However, on the user's end, `fit` multiple times isn't much more convenient that collect all tokens into a list and do a single `fit` (only takes two more lines).

Another solution would be not using `StatefulProcessorUnit` but a dedicated class.

faneshion:
We can leave the incremental preprocess to the future version. Right now, it can be satisfied for the current usage. You can pull request this code to be checked further more.

PR #219, opened 2018-07-02
Feature/vocabulary

uduse:
For issue #206


bwanglzu:
I guess this is an unfinished version?

uduse:
@bwanglzu It's finished. Any thoughts?

matchzoo/preprocessors.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
+
+    """
+
+    class IndexTerm(dict):
+        """Map index to term."""
+
+        def __missing__(self, key):
+            """Map out-of-vocabulary indices to empty string."""
+            if key == 0:
+                return ''
```
bwanglzu: I think the best thing to return is `<oov>` instead of an empty string.

matchzoo/preprocessors.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
         """
         _, input_dim = self._create_n_letters(tokens, ngram)
         self._state = {'input_dim': input_dim}
+
+
+class Vocabulary(StatefulProcessorUnit):
```
bwanglzu: can we use the same naming convention indicates that it's a `ProcessorUnit`? e.g. `VocabularyUnit`

matchzoo/preprocessors.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
+
+        def __missing__(self, key):
+            """Map out-of-vocabulary terms to index 0."""
+            return 0
+
+    def fit(self, tokens: list):
+        """Build a :class:`TermIndex` and a :clsas:`IndexTerm`."""
+        self._state['term_index'] = self.TermIndex()
+        self._state['index_term'] = self.IndexTerm()
+        terms = set(tokens)
```
bwanglzu: I'm thinking to create an independent process unit to transform terms into tokens since this might be used frequently in different models.

matchzoo/preprocessors.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
+        for index, term in enumerate(terms):
+            self._state['term_index'][term] = index + 1
+            self._state['index_term'][index + 1] = term
+
+    def transform(self, tokens: list) -> list:
```
bwanglzu: same if we have a process unit to transform tokens to terms, then we do not necessarily need to loop so many times.

matchzoo/preprocessors.py
uduse: Not necessarily. `0` could be the result of padding.

matchzoo/preprocessors.py
uduse: Sure.

matchzoo/preprocessors.py
uduse: I don't understand you.

matchzoo/preprocessors.py
uduse: Same, not sure what are you talking about.

matchzoo/preprocessors.py
uduse: I thought about separating oov and padding, but kept it simple.

matchzoo/preprocessors.py
bwanglzu:
```python
UniqueTermUnit(ProcessorUnit):
    def transform(self, tokens):
        return set(tokens)
```

uduse: I don't think it's needed.

matchzoo/preprocessors.py
bwanglzu: okay

bwanglzu:
@uduse I'm a bit confused the relationship between doctest and unit test, though we've remove some unit test, it can still covered by doctest. I personally do not know whether it's a good practice since our unt-test becomes inconsistent.

By the way, I created a folder called `preprocessors`, see pr #225 and it cause conflict..

matchzoo/preprocessor/process_units.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
+        'ABCDD'
+
+    """
+
+    class IndexTerm(dict):
+        """Map index to term."""
+
+        def __missing__(self, key):
+            """Map out-of-vocabulary indices to empty string."""
+            if key == 0:
```
pl8787: Index 0 is always used as a padding word. How about use other index, such as index 1.
0: [PAD], 1: [UNK], 2: [EOS]?

matchzoo/preprocessor/process_units.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
+                raise KeyError(key)
+
+    class TermIndex(dict):
+        """Map term to index."""
+
+        def __missing__(self, key):
+            """Map out-of-vocabulary terms to index 0."""
+            return 0
+
+    def fit(self, tokens: list):
```
pl8787: The input of this function is only one sentence in the corpus? If yes, we need consider the term already indexed.

matchzoo/preprocessor/process_units.py
```diff
@@ -228,3 +228,70 @@ def fit(self, tokens: list, ngram: int=3):
...
         self._state = {'input_dim': input_dim}
+
+
+class VocabularyUnit(StatefulProcessorUnit):
+    """
+    Vocabulary class.
+
+    Examples:
+        >>> vocab = VocabularyUnit()
+        >>> vocab.fit(['A', 'B', 'C', 'D', 'E'])
```
pl8787: vocab.fit(['hello world', 'python', 'machine learning']), is that right?

bwanglzu: the idea is `hello word` should be transformed to `hello` and `word` by TokenizeUnit.

bwanglzu: The input of `TokenizeUnit` is raw text, while input for `VocabularyUnit` is list of tokens.

matchzoo/preprocessor/process_units.py
uduse: I thought about that, but kept it simple. Anything beyond separating 0 and the others needs more detailed consideration and discussion. For example, maybe we need `OOV` but not `EOS`, maybe we need both. Users can write their own if wanted, for now.

matchzoo/preprocessor/process_units.py
uduse: For example?

PR #223, opened 2018-07-03
Python Script to download data

aneesh-joshi:
A simple script to download 
- WikiQA
- QuoraQP
- Glove

Making additions for datasets should be easy


bwanglzu:
We haven't discussed how to download such data set, but I would not create a `data` directory and use some scripts there.

Checkout [gluon-nlp](https://github.com/dmlc/gluon-nlp#dataset-loading), they provided some elegant ways to create data set.

PR #226, opened 2018-07-04
Feature/multi lang support

wqh17101:
Setup a multi-lang-doc structure


wqh17101:
Always ready. Merge when everything is ok

PR #227, opened 2018-07-04
running with TREC data set merge with 2.0

thiziri:
Is it possible to merge here?
What are your suggestions?
Thanks in advance


faneshion:
Right now, it is approximate to merge with 2.0, since the 2.0 version is not ready. By the way, we will make a dataset component to facilitate users download dataset (e.g., trec dataset) easily.

faneshion:
Thanks for all your efforts on MatchZoo. When the 2.0 version is stable, we will construct the dataset. At that time, I really hope you can help us to build the TREC dataset. Right now, I would close this PR due to the enormous amount of work to review these codes.

Issue #218, opened 2018-07-01
Implement DSSMPreprocessor


faneshion:
Instantiate a specific `Chain` for DSSM.

bwanglzu:
Process Units:
1. CollectionUnit: combine all terms in training data into Collection.
2. TokenizeUnit
3. StopRemovalUnit
4. PuncRemovalUnit
5. DigitRemovalUnit
6. LowercaseUnit
7. NgramLetterUnit, loop through all query, documents, create tri-letters, throw into `DataPack`.

faneshion:
Is there any way to save the processors? While in training, we can build a preprocessors like this:
```
preprocessor = Chain(TokenizeUnit(), Vocabulary(), ...)
```
We need to save the `preprocessor`, and load it  in future online usage to deal with test data.

bwanglzu:
Understand, the idea is to create a DSSMChain

bwanglzu:
I'm using `DSSMPreprocessor` instead of `DSSMChain`. I renamed `preprocessors` into `processor_uiits`.

bwanglzu:
dependent on #219

PR #231, opened 2018-07-07
Feature/dssm chain, add Pipeline class

bwanglzu:
For #218  , remove `base_processor_unit`, add `pipeline`.


matchzoo/engine/pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+Each model-wise preprocessor should employ a sequence of :class:`ProcessorUnit`
+and :class:`StatefulProcessorUnit` to handle input data.
+"""
+
+import typing
+
+from matchzoo import preprocessor
+
+
+class Pipeline(object):
```
uduse: I think this is the case where using a class doesn't add any functionality but complexity. I can provide a function that take a list of processor units and process them sequentially, and achieve the same functionalities as your class by simply managing the list and calling the function. Like we discussion, we don't need a class that's a fixed part of the preprocessing process (like a parent class that everyone inherits), but helper utilities that helps.

uduse: That pattern doesn't fit here.

PR #239, opened 2018-07-19
Fix a small bug(bad format) in setup.py

wqh17101:
image
there is a mistake 
the wrong format of the requirement


bwanglzu:
https://github.com/erlang/otp/wiki/writing-good-commit-messages

PR #241, opened 2018-07-19
Feature/add a pipeline base class.

bwanglzu:
For issue #218  #245 

Add a pipeline class, accept an arbitrary type of input, expect to be transformed to `DataPack` object.


matchzoo/engine/base_preprocessor.py
```diff
@@ -1,56 +1,35 @@
...
+from matchzoo import datapack
 
 
 class BasePreprocessor(metaclass=abc.ABCMeta):
-    """Abstract base class for model-wise processors."""
+    """:class:`BasePreprocessor` to input handle data."""
 
     @abc.abstractmethod
-    def _set_process_units(self):
+    def fit(self) -> typing.Callable:
```
uduse: `self` is not a callable.

Issue #245, opened 2018-07-20
[MILESTONE] DSSM integration test


bwanglzu:
After ~#238~ , ~#241~, ~#249~  ~#257~ being merged, we'll have our first 2.0 pipeline ready (DSSMModel), including pre-processing, model training,  model persistence and model evaluation.

See #268

wqh17101:
i will check the doc when these issues merged

bwanglzu:
An exmaple DSSMPreprocessor:

```python
"""DSSM Preprocessor."""

from matchzoo import engine
from matchzoo import preprocessor
from matchzoo import datapack


class DSSMPreprocessor(engine.BasePreprocessor):
    """
    DSSM preprocessor helper.

    Example:
        >>> train_inputs = [
        ...     ("beijing", "Beijing is capital of China", 1),
        ...     ("beijing", "China is in east Asia", 0),
        ...     ("beijing", "Summer in Beijing is hot.", 1)
        ... ]
        >>> dssm_preprocessor_train = DSSMPreprocessor(train_inputs)
        >>> rv_train = dssm_preprocessor_train.fit_transform()
        >>> type(rv_train)
        <class 'matchzoo.datapack.DataPack'>
        >>> fitted_parameters = rv_train.context
        >>> test_inputs = [("beijing", "I visited Beijing.", 1)]
        >>> dssm_preprocessor_test = DSSMPreprocessor(
        ...     test_inputs,
        ...     fitted_parameters)
        >>> rv_test = dssm_preprocessor_test.transform()
        >>> type(rv_test)
        <class 'matchzoo.datapack.DataPack'>

    """

    def __init__(self, inputs, context={}):
        """Initialization."""
        self.inputs = inputs
        self.context = context

    def _prepare_stateless_units(self):
        """Prepare."""
        return [
            preprocessor.TokenizeUnit(),
            preprocessor.LowercaseUnit(),
            preprocessor.PuncRemovalUnit(),
            preprocessor.StopRemovalUnit(),
            preprocessor.NgramLetterUnit()
        ]

    def _build_vocab(self):
        """Build vocabulary before fit transform."""
        vocab = []
        units = self._prepare_stateless_units()
        for left, right, label in self.inputs:
            for unit in units:
                left = unit.transform(left)
                right = unit.transform(right)
            # Extend tri-letters into vocab.
            vocab.extend(list(left.union(right)))
        return vocab

    def fit(self):
        """Fit parameters."""
[...]

PR #268, opened 2018-08-05
Feature/dssm inte test

bwanglzu:
For issue #245 
1. add integration test for dssm.
2. Resolve bug in dssm_preprocessor
3. Add utility.
4. minor changes.

### DIFF
--- a/tests/inte_test/test_dssm.py
+++ b/tests/inte_test/test_dssm.py
@@ -30,13 +30,13 @@ def inte_test_dssm():
     dssm_preprocessor = preprocessor.DSSMPreprocessor()
     processed_train = dssm_preprocessor.fit_transform(train, stage='train')
     # the dimension of dssm model is the length of tri-letters.
-    dim_triletter = processed_train.context['dim_triletter']
+    input_shapes = processed_train.context['input_shapes']
     # generator.
     generator = generators.PointGenerator(processed_train)
     X, y = generator[0]
     # Create a dssm model
     dssm_model = models.DSSMModel()
-    dssm_model.params['input_shapes'] = [(dim_triletter, ), (dim_triletter, )]
+    dssm_model.params['input_shapes'] = input_shapes
     dssm_model.guess_and_fill_missing_params()
     dssm_model.build()
     dssm_model.compile()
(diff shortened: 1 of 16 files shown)


tests/inte_test/test_dssm.py
```diff
@@ -26,11 +30,30 @@ def inte_test_dssm():
     # do pre-processing.
     dssm_preprocessor = preprocessor.DSSMPreprocessor()
```
faneshion: Here, we also need to save `dssm_preprocessor ` for prediction phase. Thus, we can load `dssm_preprocessor` for testing.

tests/inte_test/test_dssm.py
```diff
@@ -26,11 +30,30 @@ def inte_test_dssm():
...
+    dssm_model = models.DSSMModel()
+    dssm_model.params['input_shapes'] = [(dim_triletter, ), (dim_triletter, )]
+    dssm_model.guess_and_fill_missing_params()
+    dssm_model.build()
+    dssm_model.compile()
+    dssm_model.fit([X['text_left'], X['text_right']], y)
+    dssm_model.save('.tmpdir')
+
+    # testing
+    processed_test = dssm_preprocessor.fit_transform(test, stage='test')
```
faneshion: Here,  the testing phase can be implemented in an separate test function. Thus, we can also test `load dssm_preprocessors`.

tests/inte_test/test_dssm.py
```diff
@@ -0,0 +1,58 @@
...
+    processed_train = dssm_preprocessor.fit_transform(train, stage='train')
+    # the dimension of dssm model is the length of tri-letters.
+    dim_triletter = processed_train.context['dim_triletter']
+    # generator.
+    generator = generators.PointGenerator(processed_train)
+    X, y = generator[0]
+    # Create a dssm model
+    dssm_model = models.DSSMModel()
+    dssm_model.params['input_shapes'] = [(dim_triletter, ), (dim_triletter, )]
+    dssm_model.guess_and_fill_missing_params()
```
pl8787: Does `guess_and_fill_missing_params` function must be called explicitly?
I think it should be included in `build` function.

bwanglzu: @pl8787  current needed, in the future we'll write a function as "facade" to wrap these building blocks.

Issue #266, opened 2018-08-04
Implement the pair-wise generator.


faneshion:
Here, the pair-wise generator should be able to support sampling several negative samples per positive sample, rather than only one negative sample.

PR #270, opened 2018-08-05
Feature/pair generator

faneshion:
#266


faneshion:
This branch is checkout from branch feature/dssm-inte-test. It's not approximate to merge it into 2.0 directly. But you can still review this code. I will re-submit this pull request until the code is reviewed.

matchzoo/generators/pair_generator.py
```diff
@@ -0,0 +1,120 @@
...
+                for high_sample in label_info[high_label]:
+                    low_samples = []
+                    for low_label in label_list[hidx+1:]:
+                        for low_sample in label_info[low_label]:
+                            low_samples.append(low_sample)
+                    curr_pairs_num = len(low_samples) // self._num_neg
+                    for idx in range(curr_pairs_num):
+                        beg = idx * self._num_neg
+                        end = (idx + 1) * self._num_neg
+                        all_pairs.append((high_sample, low_samples[beg:end]))
```
bwanglzu: we need to re-write this part, you used 7 loops (6 for loop + 1 sorting) from Line 80- Line 92, which is extremely inefficient.

Possible solution is
1. think of another data structure to store the data.
2. Use vectorized python to replace the for loop.

matchzoo/generators/pair_generator.py
```diff
@@ -0,0 +1,120 @@
...
+        """Obtain the transformed data from :class:`DataPack`.
+
+        Note here, label is required to make pairs.
+
+        :param inputs: An instance of :class:`DataPack` to be transformed.
+        :return: the output of all the transformed inputs.
+        """
+        if 'label' not in self._column_names:
+            raise ValueError(f"label is required from {inputs} \
+                             to generate pairs.")
```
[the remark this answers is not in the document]
faneshion: Yeah, it is only used for training.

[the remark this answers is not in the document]
bwanglzu: oh got it

matchzoo/generators/point_generator.py
```diff
@@ -74,19 +75,23 @@ def _get_batch_of_transformed_samples(
         batch_y = None
         if 'label' in self.data:
             if isinstance(self._task, tasks.Ranking):
-                batch_y = self.data['label']
+                batch_y = map(int, self.data['label'])
```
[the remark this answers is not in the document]
faneshion: Yeah, that's right.

matchzoo/generators/pair_generator.py
faneshion: Here, the function `transform_data` do the same thing with `make_pair`, but use an different function name.

faneshion: @bwanglzu The pair-wise generator need to make pairs according to the labels for each query. Firstly, we need to identify all documents associated with the same query, this is what `aggregate_data` do. Then, we need to sort the documents according to the corresponding label, in this way, all samples are organized in an descending order based on labels,  such as '2 1 1 0 0 0`. Finally, we make pairs for all the possible combinations, such as '[(2, 1), (2, 1), (2, 0), (2, 0), (2, 0), (1, 0), (1, 0), (1, 0), (1, 0), (1, 0), (1, 0)]'. That's why we need so many loops.

bwanglzu: @faneshion 
1: `df.groupby("query")`
2. `df.sort("label", ascending=False)`
3. Split the dataframe by labels (say label = 0 or 1) into 2 sub-dataframes.
4. for both child dataframes, call df.sample(n, replace=True)

yangliuy: @bwanglzu So all the current data manipulations are based on pandas functions?
(1+num_neg)*k=batch_size,
df.sample(k, replace=True)
df.sample(num_neg*k, replace=True)
This part is a bit confuse. If a query has 1 positive candidate and num_neg negative candidates, we can make num_neg pairs. So should it be (num_pos*num_neg)*k=batch_size ?

bwanglzu: @yangliuy `num_pos`  alywas = 1, while `num_neg` is tunable, for example, dssm train with 1 pos together with 4 negative.

Issue #168, opened 2018-06-12
models\cdssm


bwanglzu:
implement cdssm for matchzoo 2.0.

bwanglzu:
will do in next sprint.

PR #288, opened 2018-08-14
Feature/cdssm

bwanglzu:
Add CDSSM model.

For #168

### DIFF
--- a/matchzoo/models/cdssm_model.py
+++ b/matchzoo/models/cdssm_model.py
@@ -1,0 +1,91 @@
+"""An implementation of CDSSM (CLSM) model."""
+
+from matchzoo import engine
+
+from keras import layers, Model
+
+
+class CDSSMModel(engine.BaseModel):
+    """
+    Convolutional deep structured semantic model.
+
+    Examples:
+        >>> model = CDSSMModel()
+        >>> model.guess_and_fill_missing_params()
+        >>> model.build()
+
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params()
+        params['optimizer'] = 'sgd'
+        # TODO GET TRI-LETTER DIMENSIONALITY FROM FIT-TRANSFORM AS INPUT SHAPE
+        # Dimension: (Contextual Sliding Window , NUM_TRI_LETTERS)
+        params['input_shapes'] = [(90000, 10), (90000, 50)]
+        params.add(engine.Param('w_initializer', 'glorot_normal'))
+        params.add(engine.Param('b_initializer', 'zeros'))
+        params.add(engine.Param('dim_fan_out', 128))
+        params.add(engine.Param('dim_conv', 300))
+        params.add(engine.Param('window_conv', 3))
+        params.add(engine.Param('strides', 1))
+        params.add(engine.Param('padding', 'same'))
+        params.add(engine.Param('activation_dense', 'tanh'))
+        return params
+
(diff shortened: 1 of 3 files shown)


bwanglzu:
@yangliuy the parameters can be checked through:
1. `params` property: for example, `dssm_model.params`.
2. documentation.

At the moment, all default hyper parameters = "paper". We can try our best to maintain a `consistent` set of hyper-parameters. Here the `consistent` denotes for:
1. Internal consistency: Hyper-parameters in MatchZoo models should be consistent.
2. External consistency: Hyper-parameters in MatchZoo models should be consistent with Keras and  community common naming conventions.

Besides, do you think the DSSM model implementation is correct?

logicmd:
Hi @bwanglzu , just curious, is there any issues of CDSSM model in master branch? Will CDSSM merge back to master in the future or kept as a seperate branch?
Thanks!

matchzoo/models/cdssm_model.py
```diff
@@ -0,0 +1,95 @@
...
+
+from matchzoo import engine
+
+from keras import Model
+from keras.layers import Input, Conv1D, MaxPooling1D, Dot, Dense, Flatten
+
+
+class CDSSMModel(engine.BaseModel):
+    """
+    Convolutional deep structured semantic model.
```
pl8787: We should add CDSSM paper name in the comment.
Learning Semantic Representations Using Convolutional Neural Networks for Web Search
A Latent Semantic Model with Convolutional-Pooling Structure for Information Retrieval

matchzoo/models/cdssm_model.py
```diff
@@ -0,0 +1,95 @@
...
+        >>> model.guess_and_fill_missing_params()
+        >>> model.build()
+
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params()
+        params['optimizer'] = 'sgd'
```
pl8787: the default 'sgd' optimizer could not yield a good result if we do not adjust learning rate. 
'sgd' -> 'adam' is better for default parameter.

matchzoo/models/cdssm_model.py
bwanglzu: okay

matchzoo/models/cdssm_model.py
bwanglzu: let's keep the default parameters the same as original paper, and leave the hyper-parameters tunning to the user. @pl8787

bwanglzu:
@logicmd It's for 2.0 branch only, no worries.

matchzoo/models/cdssm_model.py
faneshion: I think the default parameters is better. The better settings of parameter can be shown in the documentation of each model.

matchzoo/models/cdssm_model.py
```diff
@@ -0,0 +1,95 @@
...
+                   activation=self._params['activation_hidden'],
+                   kernel_initializer=self._params['w_initializer'],
+                   bias_initializer=self._params['b_initializer'])(x_in)
+        # Apply max pooling by take max at each dimension across
+        # all word_trigram features.
+        x = MaxPooling1D(pool_size=input_shape[1],
+                         padding=self._params['padding'])(x)
+        # Apply a none-linear transformation use a tanh layer.
+        x = Flatten()(x)
+        x_out = Dense(self._params['dim_fan_out'],
```
faneshion: The number of full-connected layers should be able to customized.  At the same time, the number of hidden nodes also need to be defined by users.

PR #292, opened 2018-08-16
add type hints for base_generator

faneshion:
#291

### DIFF
--- a/matchzoo/engine/base_generator.py
+++ b/matchzoo/engine/base_generator.py
@@ -1,6 +1,7 @@
 """Base generator."""
 
 import abc
+import typing
 import keras
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


matchzoo/engine/base_generator.py
```diff
@@ -40,7 +41,7 @@ def _set_index_array(self):
         else:
             self.index_array = np.arange(self.num_instances)
 
-    def __getitem__(self, idx):
+    def __getitem__(self, idx) -> typing.Tuple[dict, typing.Any]:
```
bwanglzu: `idx: int`

faneshion: Okay~

PR #302, opened 2018-08-18
Feature/examples

yangliuy:
The feature for adding examples of the DSSM model on the real data (WikiQA). 
To review, check https://github.com/faneshion/MatchZoo/blob/feature/examples/notebooks/wikiqa/dssm.ipynb


yangliuy:
@bwanglzu Thank you for the comments! I will update the notebook later.

bwanglzu:
we need to finish the notebook when `generator` is ready to use.

Issue #297, opened 2018-08-18
Implementation of ranking evaluations


faneshion:
The ranking evaluations include MAP, NDCG@k, P@K, MRR, and so on.

bwanglzu:
just copy -> paste [this](https://gist.github.com/bwhite/3726239) piece of code :)

PR #306, opened 2018-08-22
add metrics

bwanglzu:
for #297 ,  a copy & paste from [here](https://gist.github.com/bwhite/3726239)

Also, in this PR datapack mapping was re-written as a dict.

### DIFF
--- a/matchzoo/datapack.py
+++ b/matchzoo/datapack.py
@@ -33,10 +33,9 @@ class DataPack(object):
 
     def __init__(self,
                  data: typing.Union[list, np.ndarray],
-                 mapping: typing.Union[list, np.ndarray]=[],
+                 mapping: dict={},
                  context: dict={},
-                 columns: list=None,
-                 columns_mapping: list=None):
+                 columns: list=None):
         """
         Initialize :class:`DataPack`.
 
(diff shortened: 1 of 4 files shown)


matchzoo/datapack.py
```diff
@@ -48,11 +47,9 @@ def __init__(self,
             pre-processing stage.
         :param columns: List of column names of the :attr:`data`
             variable.
-        :param columns_mapping: List of column names of the
-            :attr:`mapping` variable.
         """
         self._dataframe = pd.DataFrame(data, columns=columns)
-        self._mapping = pd.DataFrame(mapping, columns=columns_mapping)
+        self._mapping = mapping
```
faneshion: Here, I think the `self._dataframe` should store pair information such as '(qid0, did0, label)', and `self._mapping` store the 'id: content'. In this way, we can build instances based on `self._dataframe`, then, for each id in the instance (e.g., (qid0, qid1, 1)), the content can be easily obtained from the `self._mapping` as the `self._mapping` is dict.

bwanglzu:
@faneshion I guess the best thing to do is to remove all changes related to `datapack`?

PR #317, opened 2018-08-27
Refactor/cdssm

bwanglzu:
Add support for dynamic input shape.

### DIFF
--- a/matchzoo/models/cdssm_model.py
+++ b/matchzoo/models/cdssm_model.py
@@ -3,7 +3,7 @@
 from matchzoo import engine
 
 from keras import Model
-from keras.layers import Input, Conv1D, MaxPooling1D, Dot, Dense, Flatten
+from keras.layers import Input, Conv1D, GlobalMaxPool1D, Dot, Dense
 
 
 class CDSSMModel(engine.BaseModel):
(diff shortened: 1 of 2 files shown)


matchzoo/models/cdssm_model.py
```diff
@@ -83,18 +81,15 @@ def build(self):
...
-        base_network_right = self._create_base_network(
-            input_shape=input_shape_right)
+        input_shape = self._params['input_shapes'][0]
+        base_network = self._create_base_network(
+            input_shape=input_shape)
         # Left input and right input.
-        input_left = Input(name='id_left', shape=input_shape_left)
-        input_right = Input(name='id_right', shape=input_shape_right)
+        input_left = Input(name='id_left', shape=input_shape)
+        input_right = Input(name='id_right', shape=input_shape)
```
faneshion: Are `input_left` and `input_right` both use `the input_shape[0]`?  Usually, each one has its own size.

bwanglzu: `input_left.shape = (None, 90000)`, `input_right.shape=(None, 90000)`
`None` denotes for variable input shape, better than using a `FixedLenghtProcessUnit`.

PR #321, opened 2018-08-29
new dataframe with three dataframe

faneshion:
re-design the DataPack~

### DIFF
--- a/matchzoo/datapack.py
+++ b/matchzoo/datapack.py
@@ -12,24 +12,19 @@ class DataPack(object):
     Matchzoo :class:`DataPack` data structure, store dataframe and context.
 
     Example:
-        >>> left_data = [
+        >>> left = [
         ...     ['qid1', 'query 1', 'feature 1'],
         ...     ['qid2', 'query 2', 'feature 2']
         ... ]
-        >>> right_data = [
+        >>> right = [
         ...     ['did1', 'document 1'],
         ...     ['did2', 'document 2']
         ... ]
         >>> relation = [['qid1', 'did1', 1], ['qid2', 'did2', 1]]
         >>> context = {'vocab_size': 2000}
-        >>> relation_columns = ['id_left', 'id_right', 'label']
-        >>> left_columns = ['id_left', 'left_content', 'left_feature']
-        >>> right_columns = ['id_right', 'right_content']
-        >>> relation_df = pd.DataFrame(relation, columns=relation_columns)
-        >>> left = pd.DataFrame(left_data, columns=left_columns)
-        >>> left.set_index('id_left', inplace=True)
-        >>> right = pd.DataFrame(right_data, columns=right_columns)
-        >>> right.set_index('id_right', inplace=True)
+        >>> relation_df = pd.DataFrame(relation)
+        >>> left = pd.DataFrame(left)
+        >>> right = pd.DataFrame(right)
         >>> dp = DataPack(
         ...     relation=relation_df,
         ...     left=left,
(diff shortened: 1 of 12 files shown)


matchzoo/datapack.py
```diff
@@ -13,17 +12,28 @@ class DataPack(object):
...
 
     Example:
-        >>> content = {
-        ...     'qid1':'query 1',
-        ...     'qid2':'query 2',
-        ...     'did1':'document 1',
-        ...     'did2':'document 2'
-        ... }
+        >>> left_data = [
+        ...     ['qid1', 'query 1', 'feature 1'],
```
bwanglzu: it's strange, `left_data` have `feature` while `right_data` not

matchzoo/datapack.py
```diff
@@ -36,23 +46,24 @@ class DataPack(object):
     DATA_FILENAME = 'data.dill'
 
     def __init__(self,
-                 relation: typing.Union[list, np.ndarray],
-                 content: dict,
-                 context: dict={},
-                 columns: list=None):
+                 relation: pd.DataFrame,
+                 left_data: pd.DataFrame,
+                 right_data: pd.DataFrame,
```
bwanglzu: I don't like the variable name, i.e. `left_data` and `right_data`. If I were you I would use `left` and `right` directly.

matchzoo/datapack.py
```diff
@@ -65,14 +76,30 @@ def relation(self):
...
-    @content.setter
-    def content(self, value: dict):
-        """Set the value of :attr:`content`."""
-        self._content = value
+    def left_data(self):
+        """Get :meth:`left_data` of :class:`DataPack`."""
+        return self._left_data
+
+    @left_data.setter
+    def left_data(self, value: pd.DataFrame):
```
bwanglzu: same, it's wired to use `left_data` as property, and user will never remember this property by name.

matchzoo/datapack.py
```diff
@@ -65,14 +76,30 @@ def relation(self):
...
+        """
+        self._left_data = value
+
+    @property
+    def right_data(self):
+        """Get :meth:`right_data` of :class:`DataPack`."""
+        return self._right_data
+
+    @right_data.setter
+    def right_data(self, value: pd.DataFrame):
```
bwanglzu: same

matchzoo/engine/base_preprocessor.py
```diff
@@ -89,43 +89,33 @@ def segmentation(self, inputs: list, stage: str) -> datapack.DataPack:
...
+        right_data = inputs[col_right].drop_duplicates(['id_right'])
+        right_data.set_index('id_right', inplace=True)
 
         return datapack.DataPack(relation=relation,
-                                 content=content,
-                                 columns=columns_relation)
+                                 left_data=left_data,
+                                 right_data=right_data)
 
     def _get_columns(self, stage: str) -> list:
```
bwanglzu: let's remove this function and move the logic back to `segmentation`.

matchzoo/datapack.py
faneshion: This is a real case where we only add features on query or document~

matchzoo/datapack.py
faneshion: It's a little redundant to use `left_data`. `left` is better~

tests/unit_test/models/test_models.py
```diff
@@ -17,9 +17,9 @@
 # with kwargs: (models.DenseBaselineModel, {"num_dense_units": 512})
 model_setups = [
     (models.NaiveModel, None),
-    (models.DenseBaselineModel, None),
-    (models.DSSMModel, None),
-    (models.CDSSMModel, None)
+    (models.DenseBaselineModel, None)
+#    (models.DSSMModel, None),
+#    (models.CDSSMModel, None)
```
bwanglzu: uncomment

PR #323, opened 2018-08-30
Fix a possible bug in rank_io read_data

hkvision:
line is from corpus_preprocessed, which is DocumentID followed by wordIDs
I suppose line[2:] should be changed to line[1:]? Otherwise the first word is dropped.
Feel free to correct me if I'm wrong.

### DIFF
--- a/matchzoo/utils/rank_io.py
+++ b/matchzoo/utils/rank_io.py
@@ -81,10 +81,10 @@ def read_data(filename, word_dict = None):
         line = line.strip().split()
         tid = line[0]
         if word_dict is None:
-            data[tid] = list(map(int, line[2:]))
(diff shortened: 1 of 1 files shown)


faneshion:
In the existing processing script, we have also keep the length in the corpus_preprocessed. The `line[1]` indicate the document length, thus we skip this value~

hkvision:
Thanks for your review and reply @faneshion 
But I follow this script in master branch https://github.com/faneshion/MatchZoo/blob/master/matchzoo/inputs/preprocess.py#L512 to produce corpus_preprocessed. According to the code and the result file generated, I don't see the length information is kept as the second element of a line lol...

PR #326, opened 2018-09-04
f as first argument of print() leads to unwanted file content

zenogantner:
Lines look like this:
<_io.TextIOWrapper name='./triletter_dict.txt' mode='w' encoding='UTF-8'> isf 3251

After the change:
isf 3251

### DIFF
--- a/data/WikiQA/prepare_mz_data.py
+++ b/data/WikiQA/prepare_mz_data.py
@@ -92,7 +92,7 @@ if __name__ == '__main__':
     triletter_dict = filter_triletter(triletter_stats, 5, 10000)
     with open(triletter_dict_output, 'w') as f:
         for tri_id, tric in triletter_dict.items():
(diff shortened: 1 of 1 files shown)


bwanglzu:
@zenogantner thanks for your contribution! We'll review the code asap :)

Besides, the current master branch will be completely replaced by matchzoo 2.0 once it's ready (in 2 months). If you're interested, feel free to checkout to 2.0 branch and take a look!

Best!

bwanglzu:
@zenogantner can you clarify a little bit on why do you think we shouldn't print the file?

zenogantner:
Hello @bwanglzu, see my first comment: Before my change, there were 3 columns in the created file, where the first column was the string representation of the wrapper object -- this is not the information you want to store in the file.

bwanglzu:
@zenogantner thanks. @faneshion please take a look on this one.

PR #342, opened 2018-09-10
Fix some possible import issues

hkvision:
Wondering if some of the import statements are not correct...


bwanglzu:
hi @hkvision thanks, as I mentioned in the issue, the current version of Matchzoo (including import statement) is not well written, so let's keep improving 2.0, and leave 1.0 there.

hkvision:
Sure~

Issue #298, opened 2018-08-18
Implementation of ranking loss functions 


faneshion:
The ranking loss functions include rank-hinge-loss and rank-cross-entropy.

ZizhenWang:
questions are listed below:
1. How to pass the user-defined value (like margin and neg_num) to the function? I try to define values in "losses.py" and implement a set and a get function to handle them, but it seems not safe. Another way is define a class and set these values as private,  I don't know is this way work.
2. I define a layer to slice keras tensors, to replace the Lambda layer in v1.0, I don't know is it a standard way and where to place the layer code
3. some confusing in pair-generator labels and logits shape

codes are pushed to feature/loss

ZizhenWang:
Version 2.0 pushed, it use Lambda layer instead of SliceTensor in v1.0. Which one to choose requires more discussion.

PR #345, opened 2018-09-14
Feature/loss

ZizhenWang:
#298


matchzoo/losses.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Support user-defined margin value.
+
+    :param y_true: Label.
+    :param y_pred: Predict result.
+    :return: Hinge loss computed by user-defined margin.
+    """
+    y_pos = Lambda(lambda a: a[::2, :], output_shape=(1,))(y_pred)
+    y_neg_list = []
+    for i in range(_neg_num):
```
bwanglzu: I don't like the `i` here, can we do something like:
```python
for idx in range(num_neg):
    # do something
```

matchzoo/losses.py
```diff
@@ -0,0 +1,45 @@
...
+def rank_hinge_loss(y_true, y_pred):
+    """
+    Support user-defined margin value.
+
+    :param y_true: Label.
+    :param y_pred: Predict result.
+    :return: Hinge loss computed by user-defined margin.
+    """
+    y_pos = Lambda(lambda a: a[::2, :], output_shape=(1,))(y_pred)
+    y_neg_list = []
```
bwanglzu: use the same naming conversion:

```python
y_pos = ...
y_neg = []
```

matchzoo/losses.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Support user-defined negative sample number.
+
+    :param y_true: Label.
+    :param y_pred: Predict result.
+    :return: Crossentropy loss computed by user-defined negative number.
+    """
+    y_pos_logits = Lambda(lambda a: a[::(_neg_num+1), :])(y_pred)
+    y_pos_labels = Lambda(lambda a: a[::(_neg_num+1), :])(y_true)
+    logits_list, labels_list = [y_pos_logits], [y_pos_labels]
```
bwanglzu:
```python
pos_logits, pos_labels = ...
```
or in line 35-36
```python
pos_logits = list(...)
pos_labels = list(...)
```

matchzoo/losses.py
```diff
@@ -0,0 +1,45 @@
...
+    for i in range(_neg_num):
```
bwanglzu: same, change `i` to a readable variable name.

matchzoo/losses.py
ZizhenWang: write these code in line is too long and not clean, like:
    pos_logits, pos_labels = Lambda(lambda a:
                                                            a[::(_neg_num+1), :])(y_pred),\
                                             Lambda(lambda a:
                                                            a[::(_neg_num+1), :])(y_true)
    logits_list, labels_list = [pos_logits], [pos_labels]

matchzoo/losses.py
```diff
@@ -0,0 +1,56 @@
...
+from keras import backend as K
+from keras import layers
+
+# TODO: Loss function parameters setting.
+margin = 1.0
+neg_num = 1
+
+
+def rank_hinge_loss(y_true: typing.Any,
+                    y_pred: typing.Any) -> float:
```
pl8787: I think the output value is not type `float`, but type `Tensor`.

matchzoo/losses.py
```diff
@@ -0,0 +1,56 @@
...
+    for neg_idx in range(neg_num):
+        y_neg.append(layers.Lambda(lambda a: a[(neg_idx+1)::(neg_num+1), :],
+                                   output_shape=(1,))(y_pred))
+    y_neg = K.max(K.concatenate(y_neg, axis=-1), axis=-1, keepdims=True)
+    loss = K.maximum(0., margin + y_neg - y_pos)
+    return K.mean(loss)
+
+
+def rank_crossentropy_loss(y_true: typing.Any,
+                           y_pred: typing.Any) -> float:
```
pl8787: The same

matchzoo/losses.py
```diff
@@ -0,0 +1,56 @@
+"""Self defined loss function."""
+import typing
+from keras import backend as K
+from keras import layers
+
+# TODO: Loss function parameters setting.
+margin = 1.0
+neg_num = 1
```
pl8787: The loss parameter here? Why?

ZizhenWang: It is concise to set loss parameter as global parameter in losses package, although it may leads to some permission problems.

matchzoo/losses.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def rank_hinge_loss(y_true: typing.Any,
+                    y_pred: typing.Any) -> typing.Any:
```
faneshion: `typing.Any` is not the best choice for the type of parameters. At least, you should list some possible types, such as `np.ndarray`.

ZizhenWang: The inputs and outputs type of loss function is `Tensor`, which should be represented as `typing.Any`

bwanglzu: maybe `Typing.Iterable` is better?

Issue #347, opened 2018-09-14
Add inte_test for both classification and ranking


faneshion:
Here, we need to construct 'train_ranking.txt' as well as 'test_ranking.txt' to test dssm model for both ranking and classification.

PR #348, opened 2018-09-14
test dssm with pointwise & pairwise data.

faneshion:
#347


matchzoo/generators/pair_generator.py
```diff
@@ -95,6 +95,7 @@ def transform_relation(self, relations: pd.DataFrame) -> pd.DataFrame:
         if 'label' not in relations.columns:
             raise ValueError(f"label is required from {relations} \
                              to generate pairs.")
+        relations['label'] = relations['label'].astype('int')
```
pl8787: We should use `self._task.output_dtype` instead of type of `int`.

Issue #366, opened 2018-09-28
sliding window unit


ZizhenWang:
sliding window is useful in many models, like convolution-based models

logicmd:
I thought we have used that in CDSSM, haven't we?
https://github.com/faneshion/MatchZoo/blob/ec0fef5e849f2439b08b2cf81e3192e39725eae0/examples/toy_example/config/cdssm_classify.config#L71

ZizhenWang:
@logicmd Yes! But this unit is developed for version 2.0, which we want the users code more convenient and  efficient. And it is a general unit, not only for CDSSM, but also for some other models.

PR #367, opened 2018-09-28
add sliding window unit in preprocess units.

ZizhenWang:
#366


matchzoo/preprocessor/process_units.py
```diff
@@ -346,6 +346,50 @@ def transform(
...
+        self._sliding_window = sliding_window
+
+    def transform(self, inputs: np.ndarray) -> np.ndarray:
+        """
+        Compute the result of a window sliding through input data.
+
+        :param inputs: sequential input data.
+        :return: window sliding result.
+        """
+        output = list()
```
bwanglzu: the syntax is wired.
`output = []`

matchzoo/preprocessor/process_units.py
```diff
@@ -346,6 +346,50 @@ def transform(
...
+        if len(inputs) < self._sliding_window:
+            raise ValueError(
+                'Input data can not fill user-defined window.')
```
bwanglzu: I guess the best thing to do is to use logging or warning listed of raise an Exception and use `pass` to continue

matchzoo/preprocessor/process_units.py
```diff
@@ -4,10 +4,11 @@
 import abc
 import nltk
 import typing
+import logging
 import collections
 import numpy as np
 
-
+logger = logging.getLogger(__name__)
```
bwanglzu: remove

matchzoo/preprocessor/process_units.py
```diff
@@ -346,6 +347,47 @@ def transform(
...
+
+        :param inputs: sequential input data.
+        :return: window sliding result.
+        """
+        output = []
+        while len(inputs) >= self._sliding_window:
+            output.append(np.concatenate(
+                inputs[:self._sliding_window], axis=-1))
+            inputs = inputs[1:]
+        return np.array(output)
```
bwanglzu: 1. `return` should be inside `while` loop, right? if input length less than window size, it will return `np.array(input)`.
2. Do not forget to cover the `None` case in cdssm preprocessor.

PR #382, opened 2018-10-05
factor out segmentation from base preprocessor into mixin class

bwanglzu:
This PR is important.

Put `segmentation` into `base_preprocessor` is a bit wired since the `base_preprocessor` is a high level of abstraction while `segmentation` is much more concrete.

And `segmentation` is required in multiple preprocessors, so I decided to create a Mixin class as an extra-plugin feature that can be added to multiple places using multiple inheritance, and factor out `segmentation` from `base_preprocessor`.

A more detailed description of mixin class can be found [here](https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-are-they-useful).

It might affect DSSM/CDSSM/Arc-I preprocessor implementations. @pl8787 @ZizhenWang

### DIFF
--- a/matchzoo/engine/base_preprocessor.py
+++ b/matchzoo/engine/base_preprocessor.py
@@ -2,9 +2,9 @@
 
 import abc
 import typing
-from pathlib import Path
+
 import dill
-import pandas as pd
+from pathlib import Path
 
 from matchzoo import datapack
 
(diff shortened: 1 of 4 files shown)


faneshion:
I agree with you the design of mixin for segmentation.

PR #396, opened 2018-10-10
Feature/ cdssm

bwanglzu:
add cdssm preprocessor for #328


ZizhenWang:
unfixed text length (means window number) data can not feed in cdssm model

PR #402, opened 2018-10-13
[BugFix] fix bugs in arci.

pl8787:
Fix the parameter settings from preprocessor to model.

### DIFF
--- a/matchzoo/models/arci_model.py
+++ b/matchzoo/models/arci_model.py
@@ -33,13 +33,11 @@ class ArcIModel(engine.BaseModel):
         params.add(engine.Param('right_kernel_count', [32]))
         params.add(engine.Param('right_kernel_size', [3]))
         params.add(engine.Param('activation', 'relu'))
-        params.add(engine.Param('left_pool_size', [16]))
-        params.add(engine.Param('right_pool_size', [16]))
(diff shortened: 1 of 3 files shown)


matchzoo/models/arci_model.py
```diff
@@ -52,12 +50,28 @@ def _conv_pool_block(self, input: typing.Any, kernel_count: int,
...
         """
         Build model structure.
 
         ArcI use Siamese arthitecture.
         """
+        # Check if provided embedding matrix
+        if self._params['embedding_mat'] is None:
+            self._params['embedding_mat'] = \
+               np.random.uniform(-0.2, 0.2, (self._params['vocab_size'],
+                                             self._params['embedding_dim']))
```
bwanglzu: this shouldn't exist in the model, you can use getter here to get embedding matrix, if none raise exception. And manually set embedding matrix in unit test and integration test.

matchzoo/models/arci_model.py
```diff
@@ -52,12 +50,28 @@ def _conv_pool_block(self, input: typing.Any, kernel_count: int,
         output = MaxPooling1D(pool_size=pool_size)(output)
         return output
 
+    def set_embedding_mat(self, embedding_mat: np.ndarray):
```
bwanglzu: use setter

PR #460, opened 2018-10-30
Update flake8 checking

uduse:
Create a `.flake8` file that allows ignoring certain stupid rules. e.g. "D401 First line should be in imperative mood" which cares about my tone. It's annoying even if it's 100% accurate about my tone. Also, this file can be used to ignore future annoying rules.

### DIFF
--- a/matchzoo/generators/pair_generator.py
+++ b/matchzoo/generators/pair_generator.py
@@ -136,9 +136,9 @@ class PairGenerator(engine.BaseGenerator):
         batch_x = {}
         batch_y = self._relation.iloc[trans_index, 2].values
 
-        left_vals = self._left.columns.values.tolist()
-        right_vals = self._right.columns.values.tolist()
-        columns = left_vals + right_vals + ['id_left', 'id_right']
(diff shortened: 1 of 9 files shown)


matchzoo/generators/pair_generator.py
```diff
@@ -131,12 +132,13 @@ def _get_batch_of_transformed_samples(
...
         steps = self._num_neg + 1
         for item in index_array:
-            trans_index.extend(list(range(item*steps, (item+1)*steps)))
+            trans_index.extend(list(range(item * steps, (item + 1) * steps)))
         batch_x = {}
         batch_y = self._relation.iloc[trans_index, 2].values
 
-        columns = self._left.columns.values.tolist() + \
-            self._right.columns.values.tolist() + ['id_left', 'id_right']
+        left_vals = self._left.columns.values.tolist()
```
faneshion: change `left_vals` to `left_columns`, and `right_vals` to `right_columns`.

uduse: sure

PR #461, opened 2018-10-30
Feature/update testing procedure

uduse:
`make quick` that skips slower test
`make flake` that checks flake8 only


uduse:
@bwanglzu `make quick && make flake` will do

Issue #445, opened 2018-10-25
Support Mask in generator


faneshion:
We need to provide text length for the model to support the varied-length processing.  One common practice is to add an additional mask tensor in the generator.

PR #462, opened 2018-11-02
add length in the datapack to support masking

faneshion:
#445

### DIFF
--- a/matchzoo/generators/list_generator.py
+++ b/matchzoo/generators/list_generator.py
@@ -34,8 +34,12 @@ class ListGenerator(engine.BaseGenerator):
         ...                         columns=['id_left', 'id_right', 'label'])
         >>> left = pd.DataFrame(left, columns=['id_left', 'text_left'])
         >>> left.set_index('id_left', inplace=True)
+        >>> left['length_left'] = left.apply(lambda x: len(x['text_left']),
+        ...                                  axis=1)
         >>> right = pd.DataFrame(right, columns=['id_right', 'text_right'])
         >>> right.set_index('id_right', inplace=True)
+        >>> right['length_right'] = right.apply(lambda x: len(x['text_right']),
+        ...                                     axis=1)
         >>> input = datapack.DataPack(relation=relation,
         ...                           left=left,
         ...                           right=right
@@ -52,6 +56,10 @@ class ListGenerator(engine.BaseGenerator):
         ['qid0', 'qid0', 'qid0']
         >>> x['id_right'].tolist()
         ['did0', 'did1', 'did2']
+        >>> x['length_left'].tolist()
+        [2, 2, 2]
+        >>> x['length_right'].tolist()
+        [2, 2, 2]
         >>> y.tolist()
         [0.0, 1.0, 2.0]
 
--- a/matchzoo/preprocessor/arci_preprocessor.py
+++ b/matchzoo/preprocessor/arci_preprocessor.py
@@ -148,13 +148,19 @@ class ArcIPreprocessor(engine.BasePreprocessor):
             text = row.text_left
             for unit in units:
                 text = unit.transform(text)
+            length = len(text)
             text = self._left_fixedlen_unit.transform(text)
+            length = min(length, self._fixed_length[0])
             self.datapack.left.at[idx, 'text_left'] = text
+            self.datapack.left.at[idx, 'length_left'] = length
         for idx, row in tqdm(self.datapack.right.iterrows()):
             text = row.text_right
             for unit in units:
                 text = unit.transform(text)
+            length = len(text)
             text = self._right_fixedlen_unit.transform(text)
+            length = min(length, self._fixed_length[1])
             self.datapack.right.at[idx, 'text_right'] = text
+            self.datapack.right.at[idx, 'length_right'] = length
 
         return self.datapack
--- a/matchzoo/preprocessor/cdssm_preprocessor.py
+++ b/matchzoo/preprocessor/cdssm_preprocessor.py
@@ -153,9 +153,12 @@ class CDSSMPreprocessor(engine.BasePreprocessor):
             text = [hash_unit.transform(term) for term in text]
             # flatten the text vectors
             text = list(itertools.chain(*text))
+            length = len(text)
             text = fix_unit.transform(text)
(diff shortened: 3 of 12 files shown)


notebooks/wikiqa/arci.ipynb
```diff
@@ -734,7 +734,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.3"
+   "version": "3.6.5"
```
bwanglzu: do not commit useless changes.

matchzoo/preprocessor/cdssm_preprocessor.py
```diff
@@ -164,8 +167,11 @@ def transform(
             text = [ngram_unit.transform([term]) for term in text]
             text = [hash_unit.transform(term) for term in text]
             text = list(itertools.chain(*text))
+            length = len(text)
             text = fix_unit.transform(text)
+            length = min(length, self._text_length) * self._dim_ngram
```
bwanglzu: `text_length = min(len(text), ....` and remove 170

matchzoo/preprocessor/cdssm_preprocessor.py
```diff
@@ -153,9 +153,12 @@ def transform(
             text = [hash_unit.transform(term) for term in text]
             # flatten the text vectors
             text = list(itertools.chain(*text))
+            length = len(text)
             text = fix_unit.transform(text)
+            length = min(length, self._text_length) * self._dim_ngram
```
bwanglzu: same

matchzoo/preprocessor/arci_preprocessor.py
```diff
@@ -148,13 +148,19 @@ def transform(
...
+            length = min(length, self._fixed_length[0])
             self.datapack.left.at[idx, 'text_left'] = text
+            self.datapack.left.at[idx, 'length_left'] = length
         for idx, row in tqdm(self.datapack.right.iterrows()):
             text = row.text_right
             for unit in units:
                 text = unit.transform(text)
+            length = len(text)
             text = self._right_fixedlen_unit.transform(text)
+            length = min(length, self._fixed_length[1])
```
bwanglzu: same

matchzoo/preprocessor/arci_preprocessor.py
```diff
@@ -148,13 +148,19 @@ def transform(
             text = row.text_left
             for unit in units:
                 text = unit.transform(text)
+            length = len(text)
             text = self._left_fixedlen_unit.transform(text)
+            length = min(length, self._fixed_length[0])
```
bwanglzu: same

matchzoo/preprocessor/segment.py
```diff
@@ -31,9 +31,11 @@ def segment(inputs: list, stage: str) -> datapack.DataPack:
 
     left = inputs[['id_left', 'text_left']].drop_duplicates(['id_left'])
     left.set_index('id_left', inplace=True)
+    left['length_left'] = left.apply(lambda r: len(r['text_left']), axis=1)
 
     right = inputs[['id_right', 'text_right']].drop_duplicates(['id_right'])
     right.set_index('id_right', inplace=True)
+    right['length_right'] = right.apply(lambda r: len(r['text_right']), axis=1)
```
bwanglzu: I guess you do not need to specify axis ? see [here](https://stackoverflow.com/questions/34962104/pandas-how-can-i-use-the-apply-function-for-a-single-column)

matchzoo/preprocessor/cdssm_preprocessor.py
faneshion: Here, it cannot be formalized as you suggested since the `text` have been changed in line 171. Thus, the `length` in line 170 is not equal with `len(text)` in line 172.

matchzoo/preprocessor/cdssm_preprocessor.py
faneshion: same as above~

matchzoo/preprocessor/arci_preprocessor.py
faneshion: same as above~

matchzoo/preprocessor/arci_preprocessor.py
faneshion: same as above~

matchzoo/preprocessor/cdssm_preprocessor.py
bwanglzu: rename text in line 171 then

matchzoo/preprocessor/segment.py
faneshion: It does not work to remove the `axis`. In fact, the default value of the `axis` is 0. The case you show is ok because it only modified existing values, but I have appended a new column.

notebooks/wikiqa/arci.ipynb
faneshion: I will recover this file to the previous version.

matchzoo/preprocessor/cdssm_preprocessor.py
faneshion: I think it is more appropriate to keep the name `text` to keep consistence.

notebooks/wikiqa/arci.ipynb
faneshion: All the notebooks would be modified as we will tune the parameters~

matchzoo/generators/list_generator.py
```diff
@@ -34,8 +34,12 @@ class ListGenerator(engine.BaseGenerator):
         ...                         columns=['id_left', 'id_right', 'label'])
         >>> left = pd.DataFrame(left, columns=['id_left', 'text_left'])
         >>> left.set_index('id_left', inplace=True)
+        >>> left['length_left'] = left.apply(lambda x: len(x['text_left']),
```
uduse: This `lambda` thing appeared so many times that it is better to wrap it somehow.

faneshion: The best way is to define our own dataframe and add an additional function to infer the length. That would be a big step, we can discuss it~

PR #463, opened 2018-11-02
Refactor/word hashing

bwanglzu:
for #452 not the final version, but this will resolve the memory issues.

1. Remove word hashing from dssm preprocessor into point/pair/list generator.
2. update integration test.


matchzoo/generators/list_generator.py
```diff
@@ -122,9 +126,27 @@ def _get_batch_of_transformed_samples(
         batch_x['id_left'] = id_left
         batch_x['id_right'] = id_right
 
+        if self._use_word_hashing:
+            self._hash_unit = preprocessor.WordHashingUnit(
+                self._context['term_index'])
+
         for column in self._left.columns:
+            if column == 'text_left' and self._use_word_hashing:
+                batch_x[column] = [self._hash_unit.transform(item)
```
faneshion: Can we implement a batch version of transform, such as `batch_transform`? Thus, these lines of code can be much cleaner!

PR #475, opened 2018-11-14
rewrite embedding

bwanglzu:


matchzoo/embedding.py
```diff
@@ -1,143 +1,149 @@
...
-        self._embedding_file = embedding_file
-        self._embedding_dim = 0
-        self._index_state = {}
-        self._embedding_mat = None
-
+    def __init__(self):
+        """Init."""
+        self._vocab = {}
+        self._matrix = None
+        self._dimension = 0
```
uduse: Is `_dimension` necessary? If you have `_matrix`, you always know the dimension right?

matchzoo/embedding.py
```diff
@@ -1,143 +1,149 @@
...
+        self._dimension = dimension
+        self._matrix = np.random.uniform(-random_scale,
+                                         random_scale,
+                                         (len(self._vocab), self._dimension))
+    
+    def _entity_to_vec(self, entity):
+        """"""
+        return self._matrix[self._vocab[entity]]
+
+class PretrainedEmbedding(Embedding, abc.ABC):
```
uduse: I don't think you need to inherit `abc.ABC` again.

matchzoo/embedding.py
```diff
@@ -1,143 +1,149 @@
...
-        :param embedding_file: the embeding input file path.
-        """
-        self._embedding_file = embedding_file
-        self._embedding_dim = 0
-        self._index_state = {}
-        self._embedding_mat = None
-
+    def __init__(self):
+        """Init."""
+        self._vocab = {}
```
uduse: What about accepting a `TermIndex` as an input?

matchzoo/embedding.py
```diff
@@ -1,143 +1,149 @@
...
+    def _entity_to_vec(self, entity):
+        """"""
+        return self._matrix[self._vocab[entity]]
+
+class PretrainedEmbedding(Embedding, abc.ABC):
+    """Pretrained."""
+
+    def __init__(self):
+        """"""
+        self._vocab = self._matrix.index.values
```
uduse: This is weird. You do not call `super` here, and this assignment expects `self._matrix` assignment forehand i.e. expects its child classes to assign `self._matrix` **before** calling `super`.

matchzoo/embedding.py
```diff
@@ -1,143 +1,149 @@
...
+class Word2Vec(PretrainedEmbedding):
+    """Word2Vec.
 
-            # init tht OOV word embeddings
-            for word in term_index:
-                index = term_index[word]
-                if index not in self._index_state:
-                    self._index_state[index] = 2
+    Examples:
+        >>> word2vec = Word2Vec('tests/sample/embed_10.txt')
```
uduse: I moved `tests/sample` things on `2.0-dev`. You may keep this and let me change it later, or do it yourself.

matchzoo/embedding.py
bwanglzu: yes

matchzoo/embedding.py
bwanglzu: term_index is not necessary when using pre-trained word embeddings.

matchzoo/embedding.py
bwanglzu: got it

matchzoo/embedding.py
pl8787: In this case, property `_vocab` should be an parameter of `__init__`.

uduse:
Since `RandomizedEmbedding`, `Glove` and `Word2Vec` have the same behavior after init, I suggest instead of using separate classes for them, create class methods that load them differently into the same class.

matchzoo/embedding.py
```diff
@@ -1,143 +1,68 @@
...
 
     @property
-    def index_state(self) -> dict:
-        """
-        Get word index state dictionary.
-
-        0 - [PAD] word, 1 - found embed, 2 - OOV word.
-        """
-        return self._index_state
+    def dimension(self):
```
uduse: I suggest renaming this to `output_dim` while also adding another property `input_dim` just to be consistent with namings in `BaseModel.get_default_params(with_embedding=True)`.

matchzoo/embedding.py
```diff
@@ -1,143 +1,68 @@
...
+            raise ValueError('Please load embedding.')
+        return self._matrix
+
+    @matrix.setter
+    def matrix(self, matrix_instance):
+        """Setter."""
+        self._matrix = matrix_instance
+
+    @classmethod
+    def load(cls, dir: str, embedding_type: str = 'random'):
```
uduse: `dir` is not needed for randomly initialized embeddings. I suggest separating this into to class methods, one to `load`, one to `randomly_initialize`. Also, it may be a good idea to take a random variable (e.g. scipy.rv_continueous) or an initializer (e.g. a lambda returns a random number) to initialize an embedding, but we can leave that for later.

matchzoo/embedding.py
```diff
@@ -1,143 +1,68 @@
...
+        """Load embeddings."""
+        if embedding_type == 'random':
+            self._matrix = pd.DataFrame(np.random.uniform(-random_scale,
```
uduse: `self`? `cls`? it's a `classmethod` right?

matchzoo/embedding.py
```diff
@@ -1,143 +1,68 @@
...
+                                                          random_scale,
+                                                          (len(self._vocab), self._dimension)))
```
uduse: `self._vocab` doesn't exist yet. I suggest take a matrix in `__init__`, and keep this a `classmethod`.

PR #477, opened 2018-11-16
add new function on datapack

faneshion:
add more function for datapack~


uduse:
Call `apply_on_left/right` in `apply_on_text`.

faneshion:
@uduse It cannot be implemented as you said. Because the `apply_on_left` and `apply_on_right` both return a new `pack`, and the `apply_on_text` cannot deal with the return value.

uduse:
Just added this commit https://github.com/NTMC-Community/MatchZoo/commit/ddebdde368cbe5dfb954f06a40bfa1ce0cd302b3

faneshion:
Done~

Issue #479, opened 2018-11-19
Implement DRMM_TKS


faneshion:
Re-implement DRMM_TKS for MatchZoo 2.0. The DRMM_TKS is a light version of DRMM, in which the matching histogram layer is replaced with the sorted matching top-k layer.

PR #480, opened 2018-11-19
Feature/drmm tks

faneshion:
#479

### DIFF
--- a/matchzoo/models/drmmtks_model.py
+++ b/matchzoo/models/drmmtks_model.py
@@ -1,0 +1,162 @@
+"""An implementation of DRMMTKS Model."""
+from matchzoo import engine
+import typing
+import numpy as np
+import keras.backend as K
+from keras.models import Model
+from keras.layers import Input, Embedding, Dense, Lambda, Activation, \
+    Dot, Flatten
+from keras.activations import softmax
+
+def show_tensor_info(name, input):
+    print('[Layer]: %s\t[Shape]: %s\n' % (name, input.get_shape().as_list()))
+
+class DRMMTKSModel(engine.BaseModel):
+    """
+    DRMMTKS Model.
+
+    Examples:
+        >>> model = DRMMTKSModel()
+        >>> model.guess_and_fill_missing_params()
+        >>> model.build()
+
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params()
+        params['optimizer'] = 'adam'
+        params['input_shapes'] = [(5,), (300,)]
+        params.add(engine.Param('trainable_embedding', False))
+        params.add(engine.Param('embedding_dim', 300))
+        params.add(engine.Param('vocab_size', 100))
+        params.add(engine.Param('top_k', 10))
+        params.add(engine.Param('hidden_sizes', [5, 1]))
+        params.add(engine.Param('embedding_mat', None))
+        params.add(engine.Param('embedding_random_scale', 0.2))
+        return params
+
+    @property
+    def embedding_mat(self) -> np.ndarray:
+        """Get pretrained embedding for ArcI model."""
+        # Check if provided embedding matrix
+        if self._params['embedding_mat'] is None:
+            s = self._params['embedding_random_scale']
+            self._params['embedding_mat'] = \
+                np.random.uniform(-s, s, (self._params['vocab_size'],
+                                          self._params['embedding_dim']))
+        return self._params['embedding_mat']
+
+    @embedding_mat.setter
+    def embedding_mat(self, embedding_mat: np.ndarray):
+        """
+        Set pretrained embedding for ArcI model.
+
+        :param embedding_mat: pretrained embedding in numpy format.
+        """
+        self._params['embedding_mat'] = embedding_mat
+        self._params['vocab_size'], self._params['embedding_dim'] = \
+            embedding_mat.shape
+
+    def attention_layer(self, input: typing.Any,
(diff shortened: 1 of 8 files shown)


matchzoo/models/drmmtks_model.py
```diff
@@ -0,0 +1,162 @@
...
+        params.add(engine.Param('embedding_random_scale', 0.2))
+        return params
+
+    @property
+    def embedding_mat(self) -> np.ndarray:
+        """Get pretrained embedding for ArcI model."""
+        # Check if provided embedding matrix
+        if self._params['embedding_mat'] is None:
+            s = self._params['embedding_random_scale']
+            self._params['embedding_mat'] = \
```
uduse: Do not put values like this into `self._params`. Remember that `self._params` should be able to deterministically re-create a model, and this will potentially cause problems when saving.

uduse: If you want to `remember` an embedding matrix, use something like a `self._embedding_mat` instead.

matchzoo/models/drmmtks_model.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @embedding_mat.setter
+    def embedding_mat(self, embedding_mat: np.ndarray):
+        """
+        Set pretrained embedding for ArcI model.
+
+        :param embedding_mat: pretrained embedding in numpy format.
+        """
+        self._params['embedding_mat'] = embedding_mat
+        self._params['vocab_size'], self._params['embedding_dim'] = \
```
uduse: Be really careful when saving values into `self._params`.

matchzoo/preprocessors/drmmtks_preprocessor.py
```diff
@@ -0,0 +1,94 @@
...
+        Fit pre-processing context for transformation.
+
+        :param data_pack: data_pack to be preprocessed.
+        :return: class:`DRMMTKSPreprocessor` instance.
+        """
+        units = self._default_processor_units()
+        data_pack = data_pack.apply_on_text(chain_transform(units))
+        self.vocab_unit = build_vocab(data_pack)
+
+        #self._context.update(vocab_unit.state)
```
uduse: Why not use `dict.update` instead? `index_term` should be useful.

matchzoo/preprocessors/drmmtks_preprocessor.py
```diff
@@ -0,0 +1,94 @@
...
+
+        :param data_pack: Inputs to be preprocessed.
+
+        :return: Transformed data as :class:`DataPack` object.
+        """
+        data_pack = data_pack.copy()
+        units = self._default_processor_units() + [self.vocab_unit]
+        data_pack.apply_on_text(chain_transform(units), inplace=True)
+        data_pack.apply_on_left(
+            chain_transform([self._left_fixedlength_unit]),
```
uduse:
```suggestion
            self._left_fixedlength_unit.transform,
```
Just feed the `transform` method when there's only one unit.

matchzoo/preprocessors/drmmtks_preprocessor.py
```diff
@@ -0,0 +1,94 @@
...
+            name='text_left',
+            inplace=True)
+        data_pack.apply_on_right(
+            chain_transform([self._right_fixedlength_unit]),
```
uduse: Same.

tests/inte_test/test_drmmtks.py
```diff
@@ -0,0 +1,78 @@
...
+import shutil
+
+import numpy as np
+import pytest
+
+import matchzoo as mz
+
+
+@pytest.fixture(scope='module')
+def train_data():
```
uduse: These `train_data`s in integration tests are kinda redundant. I will refactor those in the future.

matchzoo/preprocessors/drmmtks_preprocessor.py
faneshion: Got it~

matchzoo/models/drmmtks_model.py
faneshion: I agree with you to move the `_embedding_mat` out of the `_params`.

matchzoo/models/drmmtks_model.py
```diff
@@ -0,0 +1,162 @@
+"""An implementation of DRMMTKS Model."""
+from matchzoo import engine
+import typing
+import numpy as np
+import keras.backend as K
+from keras.models import Model
+from keras.layers import Input, Embedding, Dense, Lambda, Activation, \
+    Dot, Flatten
+from keras.activations import softmax
```
bwanglzu: importing conventions

matchzoo/models/drmmtks_model.py
```diff
@@ -0,0 +1,162 @@
...
+
+def show_tensor_info(name, input):
+    print('[Layer]: %s\t[Shape]: %s\n' % (name, input.get_shape().as_list()))
```
bwanglzu: you wanna user logger?

matchzoo/models/drmmtks_model.py
```diff
@@ -0,0 +1,162 @@
...
+        # shape = [B, L, 1]
+        dense_output = self.multi_layer_perceptron(matching_topk)
+        show_tensor_info("dense_output", dense_output)
+
+        # shape = [B, 1, 1]
+        dot_score = Dot(axes=[1, 1])([attention_probs, dense_output])
+        show_tensor_info("dot_score", dot_score)
+
+        flatten_score = Flatten()(dot_score)
+        show_tensor_info("flatten_score", flatten_score)
```
bwanglzu: replace

PR #488, opened 2018-11-29
Update README.md

uduse:


bwanglzu:
@uduse can you also change the description to "Facilitating the design, comparison and sharing of deep text matching models."

uduse:
> @uduse can you also change the description to "Facilitating the design, comparison and sharing of deep text matching models."

There you go.

PR #494, opened 2018-12-05
add validation callback

faneshion:


matchzoo/engine/base_model.py
```diff
@@ -205,7 +252,6 @@ def fit_generator(
         """
         return self._backend.fit_generator(
             generator=generator,
-            steps_per_epoch=len(generator),
```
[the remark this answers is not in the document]
faneshion: The default value of `steps_per_epoch` is the `len(generator)`.  In this way, users can reconfigure it as they like.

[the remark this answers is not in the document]
uduse: Then include `step_per_epoch` in `fit_generator` as a key word argument, and set it to `len(generator)` if not specified.

matchzoo/engine/base_model.py
```diff
@@ -21,6 +21,53 @@ class BaseModel(abc.ABC):
...
+            self._y = y
+            self._metrics = metrics
+            self._valid_steps = valid_steps
+            self._batch_size = batch_size
+
+        def on_epoch_end(self, epoch, logs={}):
+            if epoch % self._valid_steps:
+                return logs
+            val_x = self._x
+            val_y = self._y
```
bwanglzu: why create `val_x` and `val_y`?

matchzoo/engine/base_model.py
```diff
@@ -21,6 +21,53 @@ class BaseModel(abc.ABC):
...
+            dataframe = None
+            for metric in self._metrics:
+                if isinstance(metric, engine.BaseMetric):
+                    if dataframe is None:
+                        y_pred = self.model.predict(val_x,
+                                                    batch_size=self._batch_size
+                                                    ).reshape((-1,))
+                        data = {
+                            'id': val_x['id_left'].tolist(),
+                            'true': val_y.tolist(),
```
bwanglzu: use `label` instead of `true`?

matchzoo/engine/base_model.py
```diff
@@ -21,6 +21,53 @@ class BaseModel(abc.ABC):
...
+                        }
+                        dataframe = pd.DataFrame(data=data)
+
+                    metric_val = dataframe.groupby(by='id').apply(
+                        lambda df: metric(df['true'], df['pred'])
+                    ).mean()
+                    val_logs[str(metric)] = metric_val
+
+            print('Validation: ' + ' - '.join(
+                ['%s:%f' % (k, v) for k, v in val_logs.items()]))
```
bwanglzu: use logger instead of print

matchzoo/engine/base_model.py
uduse: It's consistent with my `evaluate`. I will change both of them into `y_true`, and `y_pred` later.

matchzoo/engine/base_model.py
uduse: We should express that these `x` and `y` are for validation, and this is one way. The other way could be naming `self._val_x` at the first place. I am ok with both.

matchzoo/engine/base_model.py
uduse: I think it would be better to worry about all the logging stuff later.

faneshion: I think we need to refactor the logger system later.

PR #509, opened 2018-12-15
Add mv_lstm into matchzoo.

zhouzhouyang520:
It should be noted that the preprocessing unit of this model is the same as that of the DSSM model, and the preprocessing unit will be updated later.

### DIFF
--- a/matchzoo/models/mvlstm_model.py
+++ b/matchzoo/models/mvlstm_model.py
@@ -23,78 +23,46 @@ class MVLSTMModel(engine.BaseModel):
     @classmethod
     def get_default_params(cls) -> engine.ParamTable:
         """:return: model default parameters."""
-        params = super().get_default_params()
+        params = super().get_default_params(with_embedding=True)
         params['optimizer'] = 'adam'
         params['input_shapes'] = [(32,), (32,)]
         params.add(engine.Param('trainable_embedding', False))
-        params.add(engine.Param('embedding_dim', 300))
-        params.add(engine.Param('embedding_input_dim', 100))
         params.add(engine.Param('vocab_size', 100))
         params.add(engine.Param('hidden_size', 32))
-        params.add(engine.Param('top_k', 10))
         params.add(engine.Param('padding', 'same'))
         params.add(engine.Param('dropout_rate', 0.0))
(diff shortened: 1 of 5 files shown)


matchzoo/models/mvlstm_model.py
```diff
@@ -0,0 +1,100 @@
...
+                              trainable=self._params['trainable_embedding'])
+        bi_lstm = Bidirectional(LSTM(self._params['hidden_size'],
+                              return_sequences=True,
+                              dropout=self._params['dropout_rate']))
+        embed_left = embedding(input_left)
+        embed_right = embedding(input_right)
+
+        rep_left = bi_lstm(embed_left)
+        rep_right = bi_lstm(embed_right)
+        cross = Dot(axes=[1, 1], normalize=False)([rep_left, rep_right])
```
bwanglzu: keras functional interface `dot` would be better.

matchzoo/models/mvlstm_model.py
```diff
@@ -0,0 +1,100 @@
...
+        >>> model.guess_and_fill_missing_params()
+        >>> model.build()
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params()
+        params['optimizer'] = 'adam'
+        params['input_shapes'] = [(32,), (32,)]
```
uduse: Why 32?

matchzoo/models/mvlstm_model.py
```diff
@@ -0,0 +1,100 @@
...
+        params.add(engine.Param('trainable_embedding', False))
+        params.add(engine.Param('embedding_dim', 300))
+        params.add(engine.Param('embedding_input_dim', 100))
+        params.add(engine.Param('vocab_size', 100))
+        params.add(engine.Param('hidden_size', 32))
+        params.add(engine.Param('top_k', 10))
+        params.add(engine.Param('padding', 'same'))
+        params.add(engine.Param('dropout_rate', 0.0))
+        params.add(engine.Param('embedding_mat', None))
```
uduse: We have embedding support now. Check ArcI or DUET.

Issue #512, opened 2018-12-15
Implement aNMM model in MatchZoo 2.0


yangliuy:
This issue is for the implementation of aNMM model in MatchZoo 2.0

yangliuy:
PR 513

PR #513, opened 2018-12-15
Feature/add anmm

yangliuy:
The pull request for adding aNMM model into MatchZoo V2.0


yangliuy:
Issue #512

sandbox.ipynb
```diff
@@ -34,7 +34,7 @@
   },
```
faneshion: Do not push changes of `sandbox`. It is still under developing.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,68 @@
...
+        # doc is [batch_size, right_text_len, bin_num]
+        query, doc = self._make_inputs()
+        embedding = self._make_embedding_layer()
+        initializer_fc = RandomUniform(minval=-0.1, maxval=0.1, seed=11)
+        initializer_gate = RandomUniform(minval=-0.01, maxval=0.01, seed=11)
+
+        q_embed = embedding(query)
+        q_w = keras.layers.Dense(
+            1, kernel_initializer=initializer_gate, use_bias=False)(q_embed)
+        q_w = keras.layers.Lambda(
```
faneshion: For the attention layer, you can refer to [drmm](https://github.com/NTMC-Community/MatchZoo/blob/feature/add_drmm/matchzoo/models/drmm.py) to implement the same function. We will extract these functions out to form the common layer.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,68 @@
...
+            lambda x: softmax(x, axis=1),
+            output_shape=(self._params['left_text_len'],)
+        )(q_w)
+        z = doc
+        z = keras.layers.Dropout(rate=self._params['dropout_rate'])(z)
+        for i in range(self._params['num_layers'] - 1):
```
faneshion: For the dense layers, you can also refer to [DRMM](https://github.com/NTMC-Community/MatchZoo/blob/feature/add_drmm/matchzoo/models/drmm.py) to simplify your code, since the dense layer has been defined in the `base_model` just like `embedding`.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,68 @@
...
+        Build model structure.
+
+        aNMM model based on bin weighting and query term attentions
+        """
+        # query is [batch_size, left_text_len]
+        # doc is [batch_size, right_text_len, bin_num]
+        query, doc = self._make_inputs()
+        embedding = self._make_embedding_layer()
+        initializer_fc = RandomUniform(minval=-0.1, maxval=0.1, seed=11)
+        initializer_gate = RandomUniform(minval=-0.01, maxval=0.01, seed=11)
```
bwanglzu: make this parameters configurable

matchzoo/models/anmm_model.py
bwanglzu: use descriptive variable name, refine `i`, `z` etc.

sandbox.ipynb
yangliuy: The modification of sandbox is tiny. The version is almost the same with the version in 2.0-dev.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,68 @@
+"""An implementation of aNMM Model."""
+
+
+import logging
+import keras
```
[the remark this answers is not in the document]
yangliuy: I will fix this.

matchzoo/models/anmm_model.py
yangliuy: I will fix this after the attention function has been included into the common layer

matchzoo/models/anmm_model.py
yangliuy: As we discussed, this will decrease the flexibility of the MLP layers. I will update this later if we want to support tuning hyper-parameters with hyper-opt.

matchzoo/models/anmm_model.py
yangliuy: I will fix this.

matchzoo/models/anmm_model.py
yangliuy: I will fix this.

yangliuy: The current implementations of models like DUET, DSSM,CDSSM, etc also use Dense layers to implement MLP. If we update all these models, the cost would be quite high and it also adds limitations on the supported MLP architectures (e.g. all activation functions and the number of nodes in each layer must be the same). We can think of other solutions if we want to support tuning hyper-parameters with hyper-opt.

.gitignore
```diff
@@ -24,3 +24,4 @@ MatchZoo.egg-info/
 notebooks/wikiqa/.ipynb_checkpoints/*
 .cache
 .tmpdir
+sandbox.py
```
uduse: Don't ignore your personal files. Just don't add them. Or, add them in your global ignore.

sandbox.ipynb
uduse: Agreed. Don't push sandbox changes here. It may cause conflicts.

matchzoo/models/dense_baseline_model.py
```diff
@@ -1,3 +1,4 @@
+
```
uduse: Why this line here?

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,70 @@
...
+        q_embed = embedding(query)
+        q_attention = keras.layers.Dense(
+            1, kernel_initializer=RandomUniform(), use_bias=False)(q_embed)
+        q_attention = keras.layers.Lambda(
+            lambda x: softmax(x, axis=1),
+            output_shape=(self._params['left_text_len'],)
+        )(q_attention)
+        d_bin = keras.layers.Dropout(
+            rate=self._params['dropout_rate'])(doc)
+        for layer_id in range(self._params['num_layers'] - 1):
```
uduse: I recommend using `enumerate` here, but your way is totally fine here.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,70 @@
...
+
+    """
+
+    @classmethod
+    def get_default_params(cls) -> engine.ParamTable:
+        """:return: model default parameters."""
+        params = super().get_default_params(with_embedding=True)
+        params.add(engine.Param('bin_num', 60))
+        params.add(engine.Param('dropout_rate', 0.1))
+        params.add(engine.Param('left_text_len', 10))
```
uduse: Isn't this a part of `input_shapes`?

matchzoo/models/__init__.py
```diff
@@ -7,8 +7,9 @@
...
 from .duet_model import DUETModel
 from .drmmtks_model import DRMMTKSModel
+from .anmm_model import ANMMModel
 from .drmm import DRMM
 
 import matchzoo
 def list_available():
-    return matchzoo.engine.BaseModel.__subclasses__()
\ No newline at end of file
+    return matchzoo.engine.BaseModel.__subclasses__()
```
uduse: Don't add this line. #515 has the same change. Both of you should be careful...

matchzoo/models/dense_baseline_model.py
yangliuy: If we remove this line, it will not pass the checking of flake8. If you run make init & make test, you will find this problem.

matchzoo/models/__init__.py
yangliuy: I will fix this.

.gitignore
yangliuy: OK. I will fix this part and do a force push after cleaning up my files.

matchzoo/models/anmm_model.py
faneshion: Maybe the `left_text_len` could be inferred from `self._params['input_shapes']`.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,70 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ANMMModel(engine.BaseModel):
+    """
+    ANMM Model.
+
+    Examples:
+        >>> model = ANMMModel()
```
faneshion: Please add the anmm-specific parameter explicitly configuration here, since the `Examples` is to show how to use `anmm`.  I have just updated existing models. You can check it.

matchzoo/models/dense_baseline_model.py
uduse: This file has been here for a long time. If anything weird happens you should first let its developer know instead of directly making a change. The problem could be your flake version, your local flake rules e.t.c. not necessarily the code. Plus, adding a new line is not the best solution.

matchzoo/models/anmm_model.py
uduse: I think it's ok to ignore the parameter setting for now. All models parameters need to be documented somehow, and I will generate part of those docs automatically later.

faneshion: That's great to automatically generate the docs.

matchzoo/models/anmm_model.py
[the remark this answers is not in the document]
yangliuy: Fixed.

matchzoo/models/anmm_model.py
yangliuy: Fixed.

matchzoo/models/anmm_model.py
yangliuy: Fixed.

.gitignore
yangliuy: Fixed.

matchzoo/models/dense_baseline_model.py
yangliuy: Fixed

sandbox.ipynb
yangliuy: Fixed.

matchzoo/models/__init__.py
yangliuy: I didn't add this line. I just added the following line in line 10 .

`from .anmm_model import ANMMModel`

matchzoo/models/anmm_model.py
yangliuy: Fixed.

yangliuy: Fixed.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,69 @@
...
+        )(q_attention)
+        d_bin = keras.layers.Dropout(
+            rate=self._params['dropout_rate'])(doc)
+        for layer_id in range(self._params['num_layers'] - 1):
+            d_bin = keras.layers.Dense(
+                self._params['hidden_sizes'][layer_id],
+                kernel_initializer=RandomUniform())(d_bin)
+            d_bin = keras.layers.Activation('tanh')(d_bin)
+        d_bin = keras.layers.Dense(
+            self._params['hidden_sizes'][self._params['num_layers'] - 1])(
```
uduse: Extract this variable for better readability.

matchzoo/models/anmm_model.py
```diff
@@ -0,0 +1,69 @@
...
+
+import logging
+
+import keras
+from keras.activations import softmax
+from keras.initializers import RandomUniform
+
+from matchzoo import engine
+
+logger = logging.getLogger(__name__)
```
uduse: loggers is not used.

Issue #436, opened 2018-10-21
Implement match-lstm.


faneshion:
Implement the basic model of match-lstm as well as the integration test.

PR #518, opened 2018-12-20
add match lstm model

faneshion:
#436


matchzoo/models/match_lstm.py
```diff
@@ -0,0 +1,78 @@
+"""Match LSTM model."""
+import keras
+import keras.backend as K
+from matchzoo import engine, preprocessors
+
+
+class MLSTMModel(engine.BaseModel):
```
uduse: Match your class name with your file name.

matchzoo/models/match_lstm.py
```diff
@@ -0,0 +1,78 @@
...
+        >>> model = MLSTMModel()
+        >>> model.guess_and_fill_missing_params(verbose=0)
+        >>> model.build()
+
+    """
+
+    @classmethod
+    def get_default_preprocessor(cls):
+        """Get default preprocessor."""
+        return preprocessors.NaivePreprocessor()
```
uduse: `BaseModel.get_default_preprocessor` already returns `NaivePreprocessor`. However, if this is just a placeholder for future development, then it's fine.

matchzoo/models/match_lstm.py
lixinsu: I have changed the model name

matchzoo/models/match_lstm.py
```diff
@@ -0,0 +1,83 @@
...
+
+    """
+
+    @classmethod
+    def get_default_params(cls):
+        """Get default parameters."""
+        params = super().get_default_params(with_embedding=True)
+        params.add(engine.Param(
+            'rnn_hidden_size', 256,
+            hyper_space=engine.hyper_spaces.quniform(low=128, high=384, q=128)
```
uduse: `q` is step size. Should be 1 here.

matchzoo/models/match_lstm.py
```diff
@@ -0,0 +1,83 @@
...
+            'fc_hidden_size', 200,
+            hyper_space=engine.hyper_spaces.quniform(
+                low=100, high=300, q=100)
+        ))
+        return params
+
+    def build(self):
+        """Build model."""
+        query, doc = self._make_inputs()
+        self.query_len = query.shape[1]
```
uduse: Do not store things in `self`. Change the signature of your attention method to take those as parameters.

matchzoo/models/match_lstm.py
lixinsu: I choose a smaller step value

matchzoo/models/match_lstm.py
lixinsu: When a tenor is passed through  LSTM layer in Keras, the seq_length dimension of the output tensor is None. And the attention function only can receive tensor as its parameter, so I change the attention as an inner function of build to get get the length information.

matchzoo/models/match_lstm.py
lixinsu: I think it is not necessary to search every possible value for the hidden size

uduse: @lixinsu It's not a grid search. `q` merely affects the granularity of sampled values.

matchzoo/models/match_lstm.py
uduse: @lixinsu Yep that is a better way. An alternative is to make it a `classmethod`.

PR #523, opened 2018-12-25
add inte_unit for drmmtks

faneshion:
Add integration test for drmmtks.


tests/inte_test/test_drmmtks.py
```diff
@@ -0,0 +1,92 @@
...
+    drmmtks_model.params['mlp_num_layers'] = 1
+    drmmtks_model.params['mlp_num_units'] = 5
+    drmmtks_model.params['mlp_num_fan_out'] = 1
+    drmmtks_model.params['mlp_activation_func'] = 'relu'
+    drmmtks_model.guess_and_fill_missing_params()
+    drmmtks_model.build()
+    drmmtks_model.compile()
+
+    x_valid, y_valid = valid_data_processed.unpack()
+    valid_eval = drmmtks_model.EvaluateOnCall(drmmtks_model, x_valid, y_valid)
```
uduse: I moved `EvaluateOnCall` into `callbacks.py`.

faneshion: I have resolved this case. However, I do not think the name `EvaluateAllMetrics` is as clear as it is. We can discuss it later~

PR #525, opened 2018-12-25
Add mvlstm into matchzoo.

zhouzhouyang520:


matchzoo/models/mvlstm.py
```diff
@@ -0,0 +1,63 @@
...
+
+from matchzoo import engine
+
+
+class MVLSTM(engine.BaseModel):
+    """
+    MVLSTM Model.
+    Examples:
+        >>> model = MVLSTMModel()
+        >>> model.guess_and_fill_missing_params()
```
faneshion: Please add the MVLSTM-specific parameter configuration just as [file](https://github.com/NTMC-Community/MatchZoo/blob/2.0-dev/matchzoo/models/drmmtks_model.py).

matchzoo/models/mvlstm.py
```diff
@@ -0,0 +1,63 @@
...
+import keras.backend as K
+
+from matchzoo import engine
+
+
+class MVLSTM(engine.BaseModel):
+    """
+    MVLSTM Model.
+    Examples:
+        >>> model = MVLSTMModel()
```
faneshion: Here should be:
```
model = MVLSTM()
```

faneshion:
Please make sure the `make test` has been successfully executed.

matchzoo/models/mvlstm.py
```diff
@@ -0,0 +1,65 @@
...
+
+        query, doc = self._make_inputs()
+
+        embedding = self._make_embedding_layer()
+        embed_query = embedding(query)
+        embed_doc = embedding(doc)
+        # Process query & document input.
+        bi_lstm = keras.layers.Bidirectional(keras.layers.LSTM(
+            self._params['hidden_size'],
+            return_sequences=True,dropout=self._params['dropout_rate']))
```
uduse:
```suggestion
            return_sequences=True, dropout=self._params['dropout_rate']))
```

matchzoo/models/mvlstm.py
```diff
@@ -0,0 +1,70 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MVLSTM(engine.BaseModel):
+    """
+    MVLSTM Model.
+
+    Examples:
+        >>> model = MVLSTM()
+        >>> model.params['mlp_num_layers']=2
```
bwanglzu: `num_mlp_layers`

matchzoo/models/mvlstm.py
```diff
@@ -0,0 +1,70 @@
...
+        >>> model.params['top_k'] = 10
```
bwanglzu: how to know the meaning of `top_k` (thinking from user's perspective)?

matchzoo/models/mvlstm.py
uduse: I guess he wanted to be consistent with those params in base model, so I think it's fine.

matchzoo/models/mvlstm.py
uduse: I'm working on automizing parameter documenting.

zhouzhouyang520: I think this parameter is related to the data. Because when I use drmm_tks model, the two data sets need different topk to achieve optimal results.

Issue #528, opened 2018-12-27
implement cdssm model to fit current vision


ZizhenWang:
the cdssm model need update to fit current frame, we should implement `CDSSM preprocessor` and update `CDSSM model`.
this issue rely on #527

PR #529, opened 2018-12-27
implement cdssm preprocessor and model

ZizhenWang:
#528

### DIFF
--- a/matchzoo/models/cdssm_model.py
+++ b/matchzoo/models/cdssm_model.py
@@ -1,9 +1,13 @@
 """An implementation of CDSSM (CLSM) model."""
+import typing
 
 from matchzoo import engine
+from matchzoo import TensorType
+from matchzoo import preprocessors
 
+from keras import backend as K
 from keras.models import Model
-from keras.layers import Input, Conv1D, GlobalMaxPool1D, Dot
+from keras.layers import Input, Conv1D, GlobalMaxPool1D, Dot, Dropout
 
 
 class CDSSMModel(engine.BaseModel):
@@ -43,9 +47,10 @@ class CDSSMModel(engine.BaseModel):
         params.add(engine.Param('strides', 1))
         params.add(engine.Param('padding', 'same'))
         params.add(engine.Param('conv_activation_func', 'tanh'))
+        params.add(engine.Param('dropout_rate', 0))
         return params
 
-    def _create_base_network(self, input_shape: tuple) -> Model:
+    def _create_base_network(self) -> typing.Callable:
         """
         Apply conv and maxpooling operation towards to each tri-letter.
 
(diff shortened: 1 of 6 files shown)


matchzoo/models/cdssm_model.py
```diff
@@ -97,3 +103,27 @@ def build(self):
         x_out = self._make_output_layer()(x)
         self._backend = Model(inputs=[input_left, input_right],
                               outputs=x_out)
+
+    @classmethod
+    def get_default_preprocessor(cls):
+        """:return: Default preprocessor."""
+        return preprocessors.CDSSMPreprocessor()
+
+    def guess_and_fill_missing_params(self, verbose=1):
```
uduse: Do your special moves, then call `super` here?

matchzoo/preprocessors/cdssm_preprocessor.py
```diff
@@ -0,0 +1,104 @@
...
+        units = self._default_processor_units()
+        units.append(processor_units.NgramLetterUnit())
+        data_pack = data_pack.apply_on_text(chain_transform(units),
+                                            verbose=verbose)
+        vocab_unit = build_vocab_unit(data_pack, verbose=verbose)
+
+        self._context['vocab_unit'] = vocab_unit
+        num_letter = len(vocab_unit.state['term_index']) + 1
+        self._context['input_shapes'] = [(self._text_len, num_letter),
+                                         (self._text_len, num_letter)]
```
uduse: What's the point of this duplication? i.e. why not just use `self._context['input_shapes'] = [self._text_len, num_letter]`? Since you customized `Input` in `build`, it's ok to customize your `input_shapes` form as well, as long as these two are consistent.

matchzoo/preprocessors/cdssm_preprocessor.py
```diff
@@ -0,0 +1,104 @@
...
+        Apply transformation on data, create `letter-ngram` representation.
+
+        :param data_pack: Inputs to be preprocessed.
+        :param verbose: Verbosity.
+
+        :return: Transformed data as :class:`DataPack` object.
+        """
+        data_pack = data_pack.copy()
+        units = self._default_processor_units()
+        units.extend([
```
uduse: `extend` is cool, but I recommend using `append` twice.

matchzoo/models/cdssm_model.py
```diff
@@ -33,59 +35,63 @@ class CDSSMModel(engine.BaseModel):
...
+        params.add(engine.Param('conv_activation_func', 'relu'))
+        params.add(engine.Param('w_initializer', 'glorot_normal'))
+        params.add(engine.Param('b_initializer', 'zeros'))
+        params.add(engine.Param('dropout_rate', 0.3))
+        params.add(engine.Param('mlp_activation_func', 'relu'))
+        params.add(engine.Param('mlp_hidden_units', [64, 32]))
         return params
 
-    def _create_base_network(self, input_shape: tuple) -> Model:
+    def _create_base_network(self) -> typing.Callable:
```
uduse: It's not required to document private methods. It's nice to have but a lot of work.

matchzoo/models/cdssm_model.py
```diff
@@ -1,29 +1,31 @@
...
+import typing
 
 from matchzoo import engine
+from matchzoo import tasks
+from matchzoo import TensorType
+from matchzoo import preprocessors
 
 from keras.models import Model
-from keras.layers import Input, Conv1D, GlobalMaxPool1D, Dot
+from keras.layers import Input, Conv1D, Dense, Dot
```
uduse: Our convention is to `import keras` then use `keras.layers.Conv1D`, `keras.layers.Dense`, e.t.c. because `matchzoo.layers` also exist and these two should not be confused.

matchzoo/models/cdssm_model.py
```diff
@@ -33,59 +35,63 @@ class CDSSMModel(engine.BaseModel):
...
         params.add(engine.Param('kernel_size', 3))
         params.add(engine.Param('strides', 1))
         params.add(engine.Param('padding', 'same'))
-        params.add(engine.Param('conv_activation_func', 'tanh'))
+        params.add(engine.Param('conv_activation_func', 'relu'))
+        params.add(engine.Param('w_initializer', 'glorot_normal'))
+        params.add(engine.Param('b_initializer', 'zeros'))
+        params.add(engine.Param('dropout_rate', 0.3))
+        params.add(engine.Param('mlp_activation_func', 'relu'))
+        params.add(engine.Param('mlp_hidden_units', [64, 32]))
```
uduse: If you understand how `matchzoo.engine.hyper_spaces` works, add at least one here, so I could test auto tunning.

matchzoo/preprocessors/cdssm_preprocessor.py
ZizhenWang: I try to simplify input shape as you recommend, but it is not iterable and can not pass `test_model.py`. I think the duplicate shape is set for different `query` and `document` text length, which not considered in `CDSSM` model. So I reset the input shapes to a list of tuple.

uduse: Try `self._context['input_shapes'] = [(self._text_len,), (num_letter,)]`. And you may also use `BaseModel._make_inputs()`.

ZizhenWang: I use `BaseModel._make_inputs()` instead and it works!

PR #533, opened 2019-01-01
[Feature] Add MatchPyramid model.

pl8787:


matchzoo/data_generator/data_generator_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+            compress_ratio_left,
+            compress_ratio_right
+        )
+        return feed_data
+
+    methodes = set(dir(generator))
+    if '_get_batch_of_transformed_samples' not in methodes:
+        return generator
+    # if 'length_left' not in generator.header?
+    setattr(generator, '_get_batch_of_transformed_samples', _wrapper)
```
uduse: This is not a simple wrapper but a really dirty hack. It changes an object's attribute implicitly and that is usually surprising. Think about using inheritance instead. Is there anything here that can't be achieved by creating a subclass of `DataGenerator`?

matchzoo/layers/dynamic_pooling_layer.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if suggestion1 != self._psize1 or suggestion2 != self._psize2:
+            raise ValueError("DynamicPooling Layer can not "
+                             "generate ({} x {}) output feature map, "
+                             "please use ({} x {} instead.)"
+                             .format(self._psize1, self._psize2,
+                                     suggestion1, suggestion2))
+
+        x_pool = K.tf.nn.max_pool(x_expand,
+                                  [1, self._msize1 / self._psize1,
```
uduse:
```suggestion
                                  [1, stride1,
```
`self._msize1 / self._psize1` is already computed.

matchzoo/layers/dynamic_pooling_layer.py
```diff
@@ -0,0 +1,92 @@
...
+        input_shape_one = input_shape[0]
+        return (None, self._psize1, self._psize2, input_shape_one[3])
+
+    def get_config(self) -> dict:
+        """Get the config dict of DynamicPoolingLayer."""
+        config = {
+            'psize1': self._psize1,
+            'psize2': self._psize2
+        }
+        base_config = super(DynamicPoolingLayer, self).get_config()
```
uduse:
```suggestion
        base_config = super().get_config()
```
Python 3.

matchzoo/layers/dynamic_pooling_layer.py
```diff
@@ -0,0 +1,92 @@
...
+        return dict(list(base_config.items()) + list(config.items()))
```
uduse:
```suggestion
        return {**base_config, **config}
```

PR #537, opened 2019-01-02
numpy doctest space

faneshion:


setup.py
```diff
@@ -15,9 +15,9 @@
     long_description = f.read()
 
 install_requires = [
-    'keras == 2.2.2',
+    'keras == 2.2.4',
```
bwanglzu: >= would be better

faneshion: ok

PR #541, opened 2019-01-03
arci tutorial

faneshion:


matchzoo/models/arci.py
```diff
@@ -0,0 +1,142 @@
+"""An implementation of ArcI Model."""
```
pl8787: Please delete the `arci_model.py`

PR #551, opened 2019-01-06
remove loss in evaluate metrics

faneshion:


pl8787:
CI did not pass. Pls fix this first.

PR #580, opened 2019-01-13
Feature/update base model default params

uduse:
I included the reasons in the commit messages.

### DIFF
--- a/matchzoo/engine/base_model.py
+++ b/matchzoo/engine/base_model.py
@@ -105,7 +105,7 @@ class BaseModel(abc.ABC):
                 desc='Should be set manually.'
             ))
             params.add(engine.Param(
-                name='embedding_trainable',
+                name='embedding_trainable', value=True,
                 desc='`True` to enable embedding layer training, '
                      '`False` to freeze embedding parameters.'
             ))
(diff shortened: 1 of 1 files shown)


matchzoo/engine/base_model.py
```diff
@@ -121,20 +116,23 @@ def get_default_params(
                      "Shouldn't be changed."
             ))
             params.add(engine.Param(
-                name='mlp_num_units',
-                desc="Number of units in first `mlp_num_layers` layers."
+                name='mlp_num_units', value=64,
```
bwanglzu: i guess 128 or 256 is more commonly used?

matchzoo/engine/base_model.py
```diff
@@ -121,20 +116,23 @@ def get_default_params(
...
             params.add(engine.Param(
-                name='mlp_num_layers',
-                desc="Number of layers of the multiple layer percetron."
+                name='mlp_num_layers', value=3,
+                desc="Number of layers of the multiple layer percetron.",
+                hyper_space=hyper_spaces.quniform(1, 6)
             ))
             params.add(engine.Param(
-                name='mlp_num_fan_out',
+                name='mlp_num_fan_out', value=32,
```
bwanglzu: 128 is better

PR #612, opened 2019-02-02
Fix the bug of attention layer in drmmtks model.

zhouzhouyang520:
Auto-tuning will be used later.


uduse:
If this is a fix on the master branch, do not include the 2.0-dev changes.

faneshion:
I think this PR is replaced by the PR 514. Thus, I would close it.

PR #614, opened 2019-02-02
Fix the mask bug of attention layer in drmm model.

zhouzhouyang520:


faneshion:
As is mentioned by @uduse, you can ignore the tutorial file~

bwanglzu:
any progress?

PR #617, opened 2019-02-11
Add data generator legacy support in v2.1.

uduse:


uduse:
CI again after #618

faneshion:
So, these data generators will be refactored?

Issue #630, opened 2019-02-21
keras metrics should be averaged in `model.evaluate` 


ZizhenWang:
current version compute keras metrics forget average

PR #631, opened 2019-02-21
#630 add mean func

ZizhenWang:


bwanglzu:
#630

bwanglzu:
but why do you think it's a bug?
image

ZizhenWang:
@bwanglzu It appears in classification task. In classification, we use categorical to compute both binary and categorical loss and metrics, which leads to
image

PR #651, opened 2019-03-17
Fix SNLI data loading index mismatch bug.

uduse:
Cause: `dropna` creates index gaps in the data frame since some of the items are removed, but `pack` generates ids consecutively and some ids might fall in the index gaps. `reset_index` fills up the index gaps and fixes the bug.


bwanglzu:
#648

PR #655, opened 2019-03-19
Feature/update wikiqa tutorial

uduse:


pl8787:
How about print some process information, when running `init.ipynb`? such as parameters have been generated.

faneshion:
ModuleNotFoundError: No module named 'tabulate'

uduse:
@faneshion Right now we don't separate dev packages (packages user do not need), so dev packages like `tabulate` are just ignored.

faneshion:
Okay~

Issue #659, opened 2019-03-24
Implement Match-SRNN


Chriskuei:
Implement Match-SRNN.

PR #660, opened 2019-03-24
[Feature] Add Match-SRNN model

Chriskuei:
#659

### DIFF
--- a/matchzoo/layers/spatial_gru.py
+++ b/matchzoo/layers/spatial_gru.py
@@ -27,7 +27,7 @@ class SpatialGRU(Layer):
         hyperbolic tangent (`tanh`). If you pass `None`, no
         activation is applied (ie. "linear" activation: `a(x) = x`).
     :param recurrent_activation: Activation function to use for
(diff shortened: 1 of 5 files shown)


matchzoo/layers/spatial_gru.py
```diff
@@ -0,0 +1,254 @@
...
+        >>> layer.build([num_batch, channel, left_len, right_len])
+
+    """
+
+    def __init__(self,
+                 units: int = 50,
+                 normalize: bool = False,
+                 init_diag: bool = False,
+                 activation: str = 'tanh',
+                 recurrent_activation: str = 'hard_sigmoid',
```
faneshion: Can we add another hyper-parameter to indicate the direction of the spatialGRU? for example, `lt` (i.e., left top) indicates the scanning from left top to right bottom, and `rb` (i.e., right bottom) indicates the scanning from right bottom to left top.

Chriskuei: Hyper-parameter `direction` has been added to MatchSRNN model.

Issue #649, opened 2019-03-16
add QuoraQP dataset


ZizhenWang:
I will add QuoraQP dataset for both ranking and classification tasks.

PR #661, opened 2019-03-25
use glue dataset as quora_qp source

ZizhenWang:
#649

### DIFF
--- a/tests/unit_test/test_datasets.py
+++ b/tests/unit_test/test_datasets.py
@@ -3,82 +3,82 @@ import pytest
 import matchzoo as mz
 
 
-@pytest.mark.slow
-def test_load_data():
-    train_data = mz.datasets.wiki_qa.load_data('train', task='ranking')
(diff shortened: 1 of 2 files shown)


tests/unit_test/test_datasets.py
```diff
@@ -62,8 +62,23 @@ def test_load_snli():
...
+        return_classes=True)
+    assert tag == [False, True]
+    assert len(dev_data) == 40431
+
+    test_data = mz.datasets.quora_qp.load_data('test')
+    assert len(test_data) == 390965
+
+    train_data = mz.datasets.quora_qp.load_data('train', 'ranking')
+    x, y = train_data.unpack()
+    assert y.shape == (363870, 1)
```
faneshion: Why not `y.shape == (363870, 2)`?

ZizhenWang: @faneshion the label of `ranking` task has shape 1

PR #662, opened 2019-03-26
Update duet tutorial

Chriskuei:


bwanglzu:
seems all metrics were decreasing during training.

Issue #672, opened 2019-04-02
Implement ESIM 2


Crysitna:
Sorry for didn't see issue #667  during implementation. Finished the ESIM model and test code already. 
Run the model on WikiQA, got map and mrr around 0.7

PR #666, opened 2019-03-27
Implement ESIM

jellying:
Implement ESIM

PR #673, opened 2019-04-02
Implemented ESIM model

Crysitna:
the implementation relates to issue #672


matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,256 @@
...
+
+        params.add(Param(
+            name='mask_value',
+            value=0,
+            desc="The value would be regarded as pad"
+        ))
+
+        return params
+
+    def compile(self, optimizer=None):
```
bwanglzu: we have a `compile` in `engine.base_model`.

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,256 @@
+"""KNRM model."""
+import keras
+import tensorflow as tf
+from keras.layers import * # Dense, Dot, Softmax, Concatenate, Embedding, LSTM, Bidirectional
```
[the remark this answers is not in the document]
bwanglzu: and it's not `KNRM` (see first line).

matchzoo/contrib/models/esim.py
Crysitna: actually wondering how can we tune the learning rate with matchzoo? I couldn't find the interface so override it to allow self-defined optimizer, will remove it

bwanglzu: @Crysitna 

```python
def get_default_params(cls)  -> ParamTable:
        params.add(Param(
            name='optimizer',
            value=keras.optimizers.SGD(lr=0.001, clipnorm=1.)
            desc="Stochastic gradient descent"
        ))
```

Do something like this?

jellying:
I run your model on WikiQA dataset but the loss descends to 0 when the second epoch starts, so does the version I implemented in pr #666 . I can't solve it yet. Would you mind sharing your running scripts or your parameters settings?

Crysitna:
@jellying  sure, shared [here](https://gist.github.com/Crysitna/1bd92cd736e111d7514a70d31d28b0a3 ). Some setup required and please refer to the README for details :)

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,241 @@
...
+        Bidirectional LSTM layer in ESIM.
+
+        :param lstm_dim: int, dimension of LSTM layer
+        :return: `keras.layers.Layer`.
+        """
+        return keras.layers.Bidirectional(
+            layer=keras.layers.LSTM(lstm_dim, return_sequences=True),
+            merge_mode='concat')
+
+    def _max(self, texts, mask) -> tf.Tensor:
```
bwanglzu: be consistent and annotate `text` and `mask` with type hints.

Or delete all type hints for internal methods.

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,241 @@
...
+        mask = self._expand_dim(mask, axis=2)
+        new_texts = keras.layers.Multiply()([texts, mask])
+
+        text_max = keras.layers.Lambda(
+            lambda x: K.max(x, axis=1),
+        )(new_texts)
+
+        return text_max
+
+    def _avg(self, texts, mask) -> tf.Tensor:
```
bwanglzu: same

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,241 @@
...
+        text_avg = keras.layers.Lambda(
+            lambda text_mask:
+                K.sum(text_mask[0], axis=1) / K.sum(text_mask[1], axis=1),
+        )([new_texts, mask])
+
+        return text_avg
+
+    def build(self):
+        """Build model."""
+
```
bwanglzu: remove empty line

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,241 @@
...
+        dropout = keras.layers.Dropout(dropout_rate)
+
+        # model
+        a, b = self._make_inputs()      # [B, T_a], [B, T_b]
+        a_mask = create_mask(a)         # [B, T_a]
+        b_mask = create_mask(b)         # [B, T_b]
+
+        ########################
+        # encoding
+        ########################
```
bwanglzu: remove `########`

One line comment: `#`

Multiline comment: 

```python
"""
This is a very long comment
comment comment line 2 comment
line 3.
"""
```

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        input_right = keras.layers.Input(
+            name='text_right',
+            shape=self._params['input_shapes'][1]
+        )
+        return [input_left, input_right]
+
+    def _make_embedding_layer(self, name='embedding') -> keras.layers.Layer:
+        """
+        Overwrite the BaseModel._make_embedding_layer to allow
```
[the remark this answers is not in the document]
bwanglzu: see [here](https://kevin.burke.dev/kevin/sphinx-interlinks/)

jellying:
Thanks a lot. I think the process that you filter the text with 0 length solves the nan promblem. I run the script esim_wikiqa.py and get the max mAP 0.6212 on test data. I wonder whether the mAP 0.702 result is from this parameter setting.

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,256 @@
...
+        a, b = self._make_inputs()      # [B, T_a], [B, T_b]
+        a_mask = create_mask(a)         # [B, T_a]
+        b_mask = create_mask(b)         # [B, T_b]
+
+        ########################
+        # encoding
+        ########################
+        a_emb = dropout(embedding(a))   # [B, T_a, E_dim]
+        b_emb = dropout(embedding(b))   # [B, T_b, E_dim]
+
```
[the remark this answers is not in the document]
Crysitna: thanks for the suggestion! let me try that

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,256 @@
...
+        mask_layer = self._make_atten_mask_layer()
+        softmax_layer = Softmax(axis=-1)
+
+        e_a = softmax_layer(mask_layer([e, _ab_mask]))          # [B, T_a, T_b]
+        e_b = softmax_layer(mask_layer([pm(e), pm(_ab_mask)]))  # [B, T_b, T_a]
+
+        # alignment (a_t = a~, b_t = b~ )
+        a_t = Dot(axes=(2, 1))([e_a, b_])   # [B, T_a, H*2]
+        b_t = Dot(axes=(2, 1))([e_b, a_])   # [B, T_b, H*2]
+
```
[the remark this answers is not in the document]
Crysitna: Oh that's actually a good point...I though the lstm would mask internally yet right, since the input it not zero the lstm might not be able to recognize them. Will fix it!

Crysitna:
> > @jellying sure, shared [here](https://gist.github.com/Crysitna/1bd92cd736e111d7514a70d31d28b0a3). Some setup required and please refer to the README for details :)
> 
> Thanks a lot. I think the process that you filter the text with 0 length solves the nan promblem. I run the script esim_wikiqa.py and get the max mAP 0.6212 on test data. I wonder how to set a better paramters to reach the mAP 0.7 mentioned in your issue #672.

That's actually wield since for the few times i run the script the map fluctuated between 0.67 to 0.71.. Let me try it again and get back to u later!
btw i did notice that the default lr hurts the performance a bit. Yet so far i'm tuning the lr by directly changing the compile() function... it's definitely not encouraged yet I still haven't find a delicate way to do it. Just you might want to try lr=4e-4 to see if the performance gets any better.

tests/inte_test/test_esim.py
```diff
@@ -0,0 +1,118 @@
...
+    return mz.pack(df)
+
+
+@pytest.fixture(scope='module')
+def train_generator(request, train_data_processed):
+    return mz.DataGenerator(train_data_processed)
+
+
+@pytest.mark.slow
+def test_duet(train_data_processed,
```
faneshion: Change the name to `test_esim`.

Crysitna:
> > @jellying sure, shared [here](https://gist.github.com/Crysitna/1bd92cd736e111d7514a70d31d28b0a3). Some setup required and please refer to the README for details :)
> 
> Thanks a lot. I think the process that you filter the text with 0 length solves the nan promblem. I run the script esim_wikiqa.py and get the max mAP 0.6212 on test data. I wonder how to set a better paramters to reach the mAP 0.7 mentioned in your issue #672.

Hi! sorry for the late reply. I'm sorry I still not sure for the reason...I actually run the same script quite a few times and the result is never lower than 0.66 (see the test result [here](https://drive.google.com/open?id=1TNNteI0imzf5Blxc5xKgPmfns5RJgNAf)). May I wonder if you use the pretrained embedding?  That's the only reason i can think for now...

matchzoo/contrib/models/esim.py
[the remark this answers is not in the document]
Crysitna: The mask are added in the new version XD

matchzoo/contrib/models/esim.py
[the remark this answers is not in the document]
Crysitna: Hi! Thanks for the advice still and I do find it theoretically reasonable, yet I tested the model with Spatial Dropout and the performance it actually downgraded a little bit. 
I upload the test log files and the comparison [here](https://drive.google.com/open?id=1TNNteI0imzf5Blxc5xKgPmfns5RJgNAf), and added the code for SpatialDropout1 version in the [gist ](https://gist.github.com/Crysitna/1bd92cd736e111d7514a70d31d28b0a3)shared last time. Feel free to comment if u find I'm not using it the right way or something!

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,238 @@
...
+        params.add(Param(
+            name='mask_value',
+            value=0,
+            desc="The value would be regarded as pad"
+        ))
+
+        return params
+
+    def _make_inputs(self) -> list:
```
uduse: This is exactly the same as the base method, why overriding it?

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,238 @@
...
+        Overwrite the :meth:`BaseModel._make_embedding_layer` to allow
+        specifiying mask_zero.
+
+        :param name: name for embedding layer, default 'embedding'
+        """
+        return keras.layers.Embedding(
+            self._params['embedding_input_dim'],
+            self._params['embedding_output_dim'],
+            trainable=self._params['embedding_trainable'],
+            mask_zero=True,
```
uduse: Only slightly different from the default one (`mask_zero`), we could update the base method to take `mask_zero` as a parameter or something.

matchzoo/contrib/models/esim.py
```diff
@@ -0,0 +1,238 @@
...
+    def _make_atten_mask_layer(self) -> keras.layers.Layer:
+        """
+        Make mask layer for attention weight matrix so that
+        each word won't pay attention to <PAD> timestep.
+        """
+        return keras.layers.Lambda(
+            lambda weight_mask: weight_mask[0] + (1.0 - weight_mask[1]) * -1e7,
+            name="atten_mask")
+
+    def _make_BiLSTM_layer(self, lstm_dim: int) -> keras.layers.Layer:
```
uduse: Only use lower case in method names.

tests/inte_test/test_esim.py
```diff
@@ -0,0 +1,118 @@
+import shutil
```
uduse: `test_models.py` is supposed to test all models, including the newly added ones. As long as the model is imported somewhere while matchzoo's loading, `test_models.py` will find it and test it. It seems there's no special setup for this model, so I guess `auto.prepare` that used by `test_models.py` will just work fine. This means it's better to not add any test file just for ESIM.

matchzoo/contrib/models/esim.py
Crysitna: oh my bad, I was trying to pass mask initially. Forgot delete this after removing the mask input.

Crysitna: > This is exactly the same as the base method, why overriding it?

oh my bad, I was trying to pass mask initially. Forgot delete this after removing the mask input.

matchzoo/contrib/models/esim.py
Crysitna: shall I send another pr for that and remove this overriding function? or just leave it for now

tests/inte_test/test_esim.py
Crysitna: oh I see, 'cause originally I saw a list of test files for each model under tests/inte_test folder and wondering why they all disappear in the new version... get it now and will remove this file.

matchzoo/contrib/models/esim.py
uduse: #688 should solve the problem, you may cherrypick it now or rebase onto it once it's merged. Also, I added you as one of the reviewers, so if it works, approve it.

PR #678, opened 2019-04-03
Feature/filtered_preprocessor

jellying:
Add a filtering process to the preprocessor.transform function, to avoid generating 0-length text in data_pack, which may cause an exception to the rnn model.

### DIFF
--- a/matchzoo/preprocessors/basic_preprocessor.py
+++ b/matchzoo/preprocessors/basic_preprocessor.py
@@ -146,4 +146,20 @@ class BasicPreprocessor(BasePreprocessor):
         data_pack.right['length_right'] = \
             data_pack.right['length_right'].apply(
                 lambda val: min(val, max_len_right))
+
+        # filter the sample with 0-length text_left or text_right in data_pack
+        keep_right_ids = set()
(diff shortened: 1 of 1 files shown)


jellying:
> A BETTER OPTIONS IS TO CREATE A PROCESSOR UNIT.

I attempt to implement it by creating a Unit like punc_removal class, but I think these units are text-level processor. The filter processor needs to be applied on whole data_pack.

bwanglzu:
@jellying can you update `FrequencyFilterUnit`? we have several processors like `BasicProcessor`, and change one of the instance doesn't really solve the issue.

jellying:
> @jellying can you update `FrequencyFilterUnit`? we have several processors like `BasicProcessor`, and change one of the instance doesn't really solve the issue.

I see. I think adding a filter function in basic class base_preprocessor can solve it. I can update it. Can I change the filter strategy and keep at least one word in each text? Or if there is a better method to avoid 0-length text exist in data_pack.

uduse:
I recommend adding a new unit for this. All units should have minimum functionalities.

bwanglzu:
me either

PR #688, opened 2019-04-08
Propagate keyword arguments to `keras.layers.Embedding`

uduse:
Propagate keyword arguments to `keras.layers.Embedding` to support things ings like masking (e.g. #673).


bwanglzu:
can you illustrate how to use it?

uduse:
@bwanglzu https://github.com/NTMC-Community/MatchZoo/blob/719ba374ff21e4232bf40e65b5a60d551b3916b4/matchzoo/contrib/models/esim.py#L62 Instead of overriding, one may simply do `_make_embedding_layer(mask_zero=True)`.

PR #693, opened 2019-04-09
seperate OOV with PAD in the vocabulary

faneshion:
It is important to separate OOV with PAD by different indices. For keras, the PAD is assumed at the index `0`. That's the basic to support the `mask_zero` in `Embedding` layer. Thus, I have placed the `PAD` at the index `0` and `OOV` at the index `1` in the `Vocabulary`.  By updating this, I have witnessed the improvement of `MVLSTM` (e,g. from 0.63 to 0.66) on WikiQA. More experiments need to be conducted on the tutorials of WikiQA.

### DIFF
--- a/matchzoo/preprocessors/units/vocabulary.py
+++ b/matchzoo/preprocessors/units/vocabulary.py
@@ -9,16 +9,16 @@ class Vocabulary(StatefulUnit):
         >>> vocab = Vocabulary()
         >>> vocab.fit(['A', 'B', 'C', 'D', 'E'])
         >>> term_index = vocab.state['term_index']
-        >>> term_index  # doctest: +SKIP
-        {'E': 2, 'C': 3, 'D': 4, 'A': 5, 'B': 6}
+        >>> term_index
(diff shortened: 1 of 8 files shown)


matchzoo/preprocessors/units/vocabulary.py
```diff
@@ -27,26 +27,14 @@ class Vocabulary(StatefulUnit):
...
         True
-        >>> vocab.transform(['C', 'A', 'OOV']) == [c_index, a_index, 1]
+        >>> vocab.transform(['C', 'A', '<OOV>']) == [c_index, a_index, 1]
         True
         >>> indices = vocab.transform(list('ABCDDZZZ'))
         >>> ' '.join(vocab.state['index_term'][i] for i in indices)
-        'A B C D D OOV OOV OOV'
+        'A B C D D <OOV> <OOV> <OOV>'
 
     """
```
uduse: I suggest adding `OOV = "<OOV>"` and `PAD = "<PAD>"` here, and use it like `self.OOV` or, from the user's perspective, `Vocabulary.OOV`.

faneshion: done~

PR #700, opened 2019-04-11
Update doc test for better test coverage and documenation.

uduse:


matchzoo/auto/tuner/callbacks/load_embedding_matrix.py
```diff
@@ -1,13 +1,39 @@
...
 
+    Example:
+
+        >>> import matchzoo as mz
+        >>> model = mz.models.ArcI()
+        >>> prpr = model.get_default_preprocessor()
+        >>> data = mz.datasets.toy.load_data()
+        >>> data = prpr.fit_transform(data, verbose=0)
+        >>> embed = mz.datasets.toy.load_embedding()
+        >>> term_index = prpr.context['vocab_unit'].state['term_index']
```
bwanglzu: you use `context` instead of `state` in #701 , correct?

uduse: These are two different PR into 2.2, will do a full replacement from state to context in the next version.

bwanglzu: @uduse ok

PR #704, opened 2019-04-11
remove invalid/empty lines from datapack

faneshion:


matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,35 @@ def drop_label(self):
...
+            >>> data_pack.drop_invalid(inplace=True)
+        """
+        if not ('length_left' in self._left and 'length_right' in self._right):
+            raise ValueError(f"`length_left` or `length_right` is missed. "
+                             f"Please call `append_text_length` in advance.")
+        invalid_left = self._left.loc[
+            self._left.length_left == 0].reset_index()
+        invalid_right = self._right.loc[
+            self._right.length_right == 0].reset_index()
+        self._left = self._left[~(self._left.index.isin(invalid_left.id_left))]
```
uduse: You are doing two inversions here. Instead, create `valid_left` for `length_left != 0`, and use it directly to slice `self._left`.

matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,35 @@ def drop_label(self):
         """
         self._relation = self._relation.drop(columns='label')
 
+    @_optional_inplace
+    def drop_invalid(self):
+        """
+        Remove rows from the data pack where the content is invalid.
+
```
uduse: You may want to explain the meaning of `invalid` here (it's length-based instead of content-based).

matchzoo/data_pack/data_pack.py
faneshion: Good idea~

matchzoo/data_pack/data_pack.py
faneshion: ok～

matchzoo/engine/base_preprocessor.py
```diff
@@ -70,6 +70,7 @@ def fit(
     def transform(
         self,
         data_pack: 'mz.DataPack',
+        drop_invalid: bool = False,
```
uduse: You shouldn't change the signature of `fit` and `transform`. Instead, you should move this parameter to `__init__`, and set the default to `True`, so you don't have to change any of the notebooks.

faneshion: It can not be added to `__init__` since the `drop_invalid` is often set to true for training set but false for validation/testing set.

ZizhenWang: how to handle the NaN  if `drop_invalid` is set to `False`

matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,34 @@ def drop_label(self):
...
+            >>> data_pack.drop_invalid(inplace=True)
+        """
+        if not ('length_left' in self._left and 'length_right' in self._right):
+            raise ValueError(f"`length_left` or `length_right` is missed. "
+                             f"Please call `append_text_length` in advance.")
+        valid_left = self._left.loc[
+            self._left.length_left != 0].reset_index()
+        valid_right = self._right.loc[
+            self._right.length_right != 0].reset_index()
+        self._left = self._left[self._left.index.isin(valid_left.id_left)]
```
bwanglzu: do you think it's correct? the `id_left` corresponds to the id in the original `left` dataframe, however, the `id` has been `reset_index` in line 314.

bwanglzu: The `id` in `left` is not equal to the `id` in `valid_left`.

faneshion: Yeah.  Since I have `reset_index` in line 315 and 316. Thus, I can use `isin(valid.id_left)`. I have tested these lines on the toy data. Maybe these codes can be simplified by drop `reset_index()`, then use `isin(valid_left.id)`?

matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,32 @@ def drop_label(self):
...
+                             f"Please call `append_text_length` in advance.")
+        valid_left = self._left.loc[self._left.length_left != 0]
+        valid_right = self._right.loc[self._right.length_right != 0]
+        self._left = self._left[self._left.index.isin(valid_left.index)]
+        self._right = self._right[self._right.index.isin(valid_right.index)]
+        self._relation = self._relation[self._relation.id_left.isin(
+            valid_left.index) & self._relation.id_right.isin(
+                valid_right.index)]
+        self._relation.reset_index(drop=True, inplace=True)
+
```
bwanglzu: the code is difficult to read, can we follow "join" -> 'process" -> "split" pipeline?

1. Join three dataframes into big one as pandas dataframe.
2. Filter columns based on conditions, such as `f = f.loc[(f['length_left'] > 0) & (f['length_right'] > 0)]` and reset index.
3. Split big dataframe into `left`, `right`, `relation` again.

I guess it's more readable and safe.

matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,32 @@ def drop_label(self):
...
+            copy. (default: `False`)
+
+        Example:
+            >>> import matchzoo as mz
+            >>> data_pack = mz.datasets.toy.load_data()
+            >>> data_pack.append_text_length(inplace=True, verbose=0)
+            >>> data_pack.drop_invalid(inplace=True)
+        """
+        if not ('length_left' in self._left and 'length_right' in self._right):
+            raise ValueError(f"`length_left` or `length_right` is missed. "
```
bwanglzu: is missing

matchzoo/data_pack/data_pack.py
uduse: I think the code now is acceptable.

PR #719, opened 2019-04-18
ignore notebook from code percentage calculate

bwanglzu:
turn matchzoo into a python project


uduse:
Are you sure this works? This is the view I got on this commit.

image

bwanglzu:
@uduse see [this link](https://stackoverflow.com/questions/29932342/is-there-a-way-to-ignore-calculating-language-statistics-for-a-directory-on-gith), guess it will only happens on master branch

uduse:
image

I believe you got the path wrong. I tested this on a fork and it works.

Issue #717, opened 2019-04-15
implement BPE preprocessor


ZizhenWang:
BPE is used many NLP tasks as machine translation, generation and pre-training, we will implement a BPE processor to support these requirements.

bwanglzu:
but why do you think it's related to text matching?

ZizhenWang:
It is basic tokenizer of Bert.

bwanglzu:
if it's a tokenizer, then should be designed as a processor unit.

sth4k:
I think it's more to solve the OOV (out-of-vocabulary) problem. Currently the vocabulary is only built from training data and thus maybe limited by how you choose your training data.

PR #722, opened 2019-04-20
Add berttokenize unit.

jellying:
#717


bwanglzu:
会破坏整个项目的代码结构。

jellying:
> 会破坏整个项目的代码结构。

最关键的问题想用bert就要加载bert自身的vocab，而不是由vocabulary这个unit生成，需要想办法设计一下怎么兼容bert的词表和预处理。

matchzoo/preprocessors/bert_preprocessor.py
```diff
@@ -0,0 +1,141 @@
...
+        Bert-base Model preprocessor.
+
+        TODO: doc here.
+
+
+        Example:
+            >>> import matchzoo as mz
+            >>> train_data = mz.datasets.toy.load_data()
+            >>> test_data = mz.datasets.toy.load_data(stage='test')
+            >>> # the argument 'bert_vocab_path' must feed the bert vocab path
```
bwanglzu: 1. The first letter in the comment need to be upper case.
2. The argument `bert_vocab_path` must feed the bert vocab path, you mentioned **must**, then please check the argument in the function and remove default value.
3. It's better for us to unify the function-level indent, i,e,

```python
def __init__(
    self,
    fixed_length_left: int = 30,
    ...
)
```

4. Remove TODO.

matchzoo/preprocessors/units/tokenize.py
```diff
@@ -15,3 +17,104 @@ def transform(self, input_: str) -> list:
...
+        output = []
+        for char in input_:
+            cp = ord(char)
+            if is_chinese_char(cp):
+                output.append(" ")
+                output.append(char)
+                output.append(" ")
+            else:
+                output.append(char)
+        return "".join(output)
```
bwanglzu: Do you think it's correct? Let's say you give two Chinese chars: 'a' and 'b'

`output = [' ', 'a', ' ', ' ', 'b', ' ']`

And you join the result, it will be two blanks instead of one blank as described in your `:return:` section.

matchzoo/preprocessors/bert_preprocessor.py
```diff
@@ -0,0 +1,141 @@
...
+            data_pack.right['length_right'].apply(
+                lambda val: min(val, max_len_right))
+        return data_pack
+
+    @classmethod
+    def _default_units(cls) -> list:
+        """Prepare needed process units."""
+        return [
+            units.BertClean(),
+            units.BasicTokenize()
```
bwanglzu: you set `chinese_version` is `True` in `__init__`, why not use `ChineseTokenize` in default units?

matchzoo/utils/bert_utils.py
```diff
@@ -0,0 +1,104 @@
+import unicodedata
+
+
+def is_whitespace(char):
+    """Checks whether `chars` is a whitespace character."""
+    # \t, \n, and \r are technically contorl characters but we treat them
+    # as whitespace since they are generally considered as such.
+    if char == " " or char == "\t" or char == "\n" or char == "\r":
```
bwanglzu:
```python
or ... or ... or unicodedata.category(char) == 'Zs':
    return True
return False
```

And mroeve 11 - 12, more readable.

matchzoo/utils/bert_utils.py
```diff
@@ -0,0 +1,104 @@
...
+def is_control(char):
+    """Checks whether `chars` is a control character."""
+    # These are technically control characters but we count them as whitespace
+    # characters.
+    if char == "\t" or char == "\n" or char == "\r":
+        return False
+    cat = unicodedata.category(char)
+    if cat in ("Cc", "Cf"):
+        return True
+    return False
```
bwanglzu:
```suggestion
if cat in ['Cc', 'Cf']:
    return True
return False
```

matchzoo/utils/bert_utils.py
```diff
@@ -0,0 +1,104 @@
...
+def is_punctuation(char):
+    """Checks whether `chars` is a punctuation character."""
+    cp = ord(char)
+    # We treat all non-letter/number ASCII as punctuation.
+    # Characters such as "^", "$", and "`" are not in the Unicode
+    # Punctuation class but we treat them as punctuation anyways, for
+    # consistency.
+    condition = (33 <= cp <= 47) or (58 <= cp <= 64) or \
+                (91 <= cp <= 96) or (123 <= cp <= 126)
+    if condition:
```
bwanglzu:
```suggestion
    if condition or cat.startswidth('P'):
        return True
    return False
```

matchzoo/preprocessors/units/tokenize.py
```diff
@@ -15,3 +17,104 @@ def transform(self, input_: str) -> list:
...
+
+        :return tokens: tokenized tokens as a list.
+        """
+        orig_tokens = whitespace_tokenize(input_)
+        split_tokens = []
+        for token in orig_tokens:
+            split_tokens.extend(run_split_on_punc(token))
+
+        output_tokens = whitespace_tokenize(" ".join(split_tokens))
+        return output_tokens
```
bwanglzu: What's the difference between this one and the NLTK one?

matchzoo/preprocessors/units/tokenize.py
```diff
@@ -15,3 +17,104 @@ def transform(self, input_: str) -> list:
...
+            chars = list(token)
+            if len(chars) > self.max_input_chars_per_word:
+                output_tokens.append(self.unk_token)
+                continue
+
+            is_bad = False
+            start = 0
+            sub_tokens = []
+            while start < len(chars):
+                end = len(chars)
```
bwanglzu: why put this line inside of the while loop?

matchzoo/preprocessors/units/tokenize.py
```diff
@@ -15,3 +17,104 @@ def transform(self, input_: str) -> list:
...
+        :return tokens: A list of wordpiece tokens.
+        """
+        output_tokens = []
+        for token in input_:
+            chars = list(token)
+            if len(chars) > self.max_input_chars_per_word:
+                output_tokens.append(self.unk_token)
+                continue
+
+            is_bad = False
```
bwanglzu: can you rename this variable, difficult to understand.

matchzoo/preprocessors/units/tokenize.py
jellying: The unit ``ChineseTokenize`` just adds whitespaces around Chinese chars to solve the cases like ``input_ = "abcd"``, so the return type is str. The redundant whitespaces will be cleaned in the next unit ``BasicTokenize`` and I will change the comment as ``return output: text with at least one blank ...``. Maybe there is a proper name for this unit.

matchzoo/preprocessors/bert_preprocessor.py
jellying: set the ``chinese_version`` is ``False`` for users'  different demands.

matchzoo/utils/bert_utils.py
jellying: Because of the codestyle constraints about the if condition format, I change it as follow:
``return (char == " ") or \
           (char == "\t") or \
           (char == "\n") or \
           (char == "\r") or \
           (unicodedata.category(char) == "Zs")``
With the constraint of length, each token `\` follows a break line and the parentheses `()` is added to avoid codestyle error.

matchzoo/preprocessors/units/tokenize.py
jellying: At line 110, if the prefix is not matched in vocab, the end index shrinks by step to find the longest prefix and tokenize.

matchzoo/preprocessors/units/tokenize.py
jellying: rename it as `failed_match`?

bwanglzu: @jellying the logic of this function is a bit complicate, I need to read it through to better understand it.. So I don't know what you wanna do here.

matchzoo/preprocessors/units/tokenize.py
bwanglzu: but `len(chars)` never changed, right?

matchzoo/utils/bert_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    return ["".join(x) for x in output]
+
+
+def whitespace_tokenize(text):
+    """Runs basic whitespace cleaning and splitting on a piece of text."""
+    text = text.strip()
+    if not text:
+        return []
+    tokens = text.split()
+    return tokens
```
bwanglzu: image

`if not text` is not necessarily needed.

matchzoo/preprocessors/units/tokenize.py
jellying: In `nltk` package, `TreebankWordTokenizer` is used and words like `can't` are tokenized as `ca` and `n't`. But in bert-based model, this process should be executed by the `WordPieceTokenize` to avoid improper tokenization results.

matchzoo/preprocessors/units/tokenize.py
jellying: Yes, I set a variable to store it.

matchzoo/preprocessors/units/tokenize.py
jellying: For example, we need to tokenize the word 'standard' and vocab = ['stand', 'dard'], we first match the prefix 'stand', but residual text  'ard' is not existed in vocab and 'a', 'r', 'd' neither. Without proper longest pieces in vocab to construct the target word, so the `is_bad` set `True` and we mark it as '[unk]'.

bwanglzu: @jellying then it should be something like `unknow_suffix`

matchzoo/preprocessors/units/tokenize.py
bwanglzu: they why not remove line 40?

matchzoo/utils/bert_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+def run_strip_accents(text):
+    """Strips accents from a piece of text."""
+    text = unicodedata.normalize("NFD", text)
+    output = []
+    for char in text:
+        cat = unicodedata.category(char)
+        if cat == "Mn":
+            continue
+        output.append(char)
```
bwanglzu:
```suggestion
output = [char for char in text if not unicodedata.category(char) == 'Mn']
```

matchzoo/preprocessors/units/vocabulary.py
```diff
@@ -67,3 +67,44 @@ def fit(self, tokens: list):
...
+
+    """
+
+    def __init__(self, pad_value: str = '[PAD]', oov_value: str = '[UNK]'):
+        """Vocabulary unit initializer."""
+        super().__init__()
+        self._pad = pad_value
+        self._oov = oov_value
+        self._context['term_index'] = self.TermIndex()
+        self._context['index_term'] = dict()
```
bwanglzu: `
```suggestion
        self._context['index_term'] = {}
```

matchzoo/utils/bert_utils.py
bwanglzu: can you see this one?

matchzoo/preprocessors/units/__init__.py
```diff
@@ -13,6 +13,12 @@
 from .tokenize import Tokenize
 from .vocabulary import Vocabulary
 from .word_hashing import WordHashing
+from .bert_clean import BertClean
+from .bert_clean import StripAccent
+from.tokenize import ChineseTokenize
```
bwanglzu: no space

matchzoo/preprocessors/units/tokenize.py
```diff
@@ -15,3 +17,106 @@ def transform(self, input_: str) -> list:
...
+    def transform(self, input_: list) -> list:
+        """
+        Tokenizes a piece of text into its word pieces.
+
+        This uses a greedy longest-match-first algorithm to perform
+         tokenization using the given vocabulary.
+
+        For example:
+          input = "unaffable"
+          output = ["un", "##aff", "##able"]
```
bwanglzu: change this with doctest.

bwanglzu:
@uduse can you review?

Issue #728, opened 2019-04-25
Optimizers with different learning rate multipliers


Chriskuei:
Default keras optimizers have a global learning rate. But sometimes setting different learning rates for different layers may result in better performance.

PR #729, opened 2019-04-25
[Feature] Optimizers with LR multipliers

Chriskuei:
This PR implements optimizers with learning rate multipliers, which can be used to apply different learning rates for different layers. For example:

```python
optimizer = MultiSGD(multipliers={'dense_1': 0.8})
```

In this example, parameter `multipliers` indicates that layers whose name contain `dense_1` have a learning rate multiplier of 0.8 with respect to global learning rate.

### Related issue
#728

### DIFF
--- a/matchzoo/__init__.py
+++ b/matchzoo/__init__.py
@@ -33,6 +33,7 @@ from . import datasets
 from . import layers
 from . import auto
 from . import contrib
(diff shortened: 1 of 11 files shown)


bwanglzu:
should be a pr to keras rather than matchzoo.

PR #730, opened 2019-04-27
add DIIN model

caiyinqiong:


matchzoo/contrib/layers/encoding_layer.py
```diff
@@ -0,0 +1,122 @@
+"""An implementation of EncodingModule for DIIN model."""
```
ZizhenWang: I think `encoding layer` is required by almost every model, and the structures are not consistent. If it is difficult to write this layer in `DIIN` model, you'd better use another name.

matchzoo/contrib/models/diin.py
```diff
@@ -0,0 +1,310 @@
...
+        params.add(Param(name='layers_per_dense_block', value=8,
+                         desc="The number of convolution layers in dense "
+                              "block."))
+        params.add(Param(name='growth_rate', value=20,
+                         desc="The filter size of each convolution layer in "
+                              "dense block."))
+        params.add(Param(name='transition_scale_down_ratio', value=0.5,
+                         desc="The channel scale down ratio of the "
+                              "convolution layer in transition block."))
+        return params
```
uduse: Assign at least one hyper-space so auto tuner will work.

matchzoo/contrib/models/diin.py
```diff
@@ -0,0 +1,310 @@
...
+        #   B = batch size (number of sequences)
+        #   D = word embedding size
+        #   L = 'input_left' sequence length
+        #   R = 'input_right' sequence length
+        #   C = fixed word length
+
+        # Left text and right text.
+        # shape = [B, L]
+        # shape = [B, R]
+        text_left, text_right = self._make_inputs()[0:2]
```
uduse: Shouldn't call this multiple times. Do a `inputs = self._make_inputs()` then slice `inputs`.

PR #765, opened 2019-07-12
add drop_invlid

faneshion:


requirements.txt
```diff
@@ -15,3 +15,4 @@ pytest >= 3.7.4
 pytest-cov >= 2.4.0
 flake8 >= 3.6.0
 flake8_docstrings >= 1.3.0
+pydocstyle == 2.1
```
uduse: Why?

matchzoo/data_pack/data_pack.py
```diff
@@ -294,6 +294,32 @@ def drop_label(self):
...
+            >>> data_pack.drop_invalid(inplace=True)
+        """
+        if not ('length_left' in self._left and 'length_right' in self._right):
+            raise ValueError(f"`lenght_left` or `length_right` is missing. "
+                             f"Please call `append_text_length` in advance.")
+        valid_left = self._left.loc[self._left.length_left != 0]
+        valid_right = self._right.loc[self._right.length_right != 0]
+        self._left = self._left[self._left.index.isin(valid_left.index)]
+        self._right = self._right[self._right.index.isin(valid_right.index)]
+        self._relation = self._relation[self._relation.id_left.isin(
```
uduse: I recommend extracting this inner part as a local variable.

requirements.txt
faneshion: This is for the cron CI. I have tested in the master branch.

PR #779, opened 2019-08-07
Update rank_cross_entropy_loss to avoid NANs

jibrilfrej:
Added an epsilon in the log of the rank cross entropy loss


uduse:
:tada::tada::tada:

PR #795, opened 2019-10-12
Switch to tf.keras [WIP]

matthew-z:
DONE:
* switch to tf.keras and remove keras dependency
* upgrade ubuntu dist from 14.04 to 16.04 in CI, as tensorflow 2.0 requires at least 16.04
* upgrade setuptools in CI
* add keywords param in `model.compile` to allow users to disable some experimental tf2.0 features 
* add pytest-xdist plugin to run each test in separate processes using fork (note that fork does not work in Windows)
* upgraded pytest to 4.4, as required by pytest-xdist

TODO:
* BiMPM and MultiPerspectiveLayer: need refactoring to avoid initializing layers in `call`, which causes errors with the tf 2.0 default eager mode. 
* MatchLSTM:  I think tf.layers.LSTM + model.fit/evaluate still have some bugs to fix, see https://github.com/tensorflow/tensorflow/issues/33258. Currently I set `experimental_run_tf_function` to false as a workaround. However, I still encountered some errors in CI from time to time (randomly happened), while local test is usually OK....

~PS: since there is no 2.3 dev branch, the current PR base is the master branch.~ changed the base to 2.3-dev


bwanglzu:
@matthew-z I cleaned up all the archived branches, please send future pr to `2.3-dev`.

